PR #24, opened 2014-08-26
Removed ionic.zip

### LINKED ISSUE #3: Nuke Ionic.Zip
We can probably inline this and use the framework for Zips as @justinvp says

distantcam:
The only major change is there's no bzip2 compression in the framework, so I swapped to gzip. Fixes #3 
## TODO:
- [ ] Fix the broken tests


rjvdboon:
This PR can be closed I guess, there is no dependency on Ionic.Zip anymore.

Thieum:
superseded and done in #803

PR #73, opened 2014-09-28
Sample app for testing

christianrondeau:
This PR is far from being complete, but I wanted to run this through you before putting any more time on it.

Here's the idea:
- Create an _extremely_ simple app to install and uninstall
- Make that app log very clear information on it's current version, and what is being executed exactly
- Run an install, an uninstall and verify
  - That the files are indeed present in the expected location
  - That the log file contains exactly what is supposed to be there
- Run an update
  - from multiple variations of the Releases directory
  - From the file system as well as an IIS mapping (including a Web.config to make it easy to setup)

So, either using xunit or just using a PowerShell script, we'll be able to test most edge cases in integration tests.

This is better than what I had to do manually, which is manually try cases with my own app.

It will also help documentation by providing a real life example people can look for when trying to understand how they should integrate.

What do you think?

### DIFF
--- a/Squirrel.sln
+++ b/Squirrel.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.30501.0
+VisualStudioVersion = 12.0.30723.0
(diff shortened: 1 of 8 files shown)


moser96:
I think it would be nice if you add the code for **querying the version and the release notes**, so other users can lookup this code in your demo.

PR #154, opened 2014-12-07
Releasify - Change the setup icon 

BarryThePenguin:
An attempt at changing the setup.exe icon similar to how the zip data is changed. This is my first venture into pinvokes. Any pointers? :joy: :facepunch:


paulcbetts:
I fixed this up in #183 the lazy way - thanks for having a go at it!

Issue #137, opened 2014-11-24
ui is inadequate


mark-hahn:
Installing the github atom editor takes up to a minute.  Everything looks broken because a graphic just sits there.  
- no progress indicator
- No indication what is happening
- No window border which causes it to get lost when overlapping windows
- No icon to click on the taskbar

You paradigm of the install being instantaneous sounds good in theory but it isn't feasible.

PR #184, opened 2015-01-12
Disable animated GIF window from showing up in the Taskbar

paulcbetts:
This PR removes the installation window from the Taskbar and makes it Always On Top for the first 5 seconds (then allows it to be backgrounded)

### DIFF
--- a/src/Update/AnimatedGifWindow.cs
+++ b/src/Update/AnimatedGifWindow.cs
@@ -63,14 +63,10 @@ namespace Squirrel.Update
                 wnd = new AnimatedGifWindow();
                 wnd.Show();
 
-                var dt = new DispatcherTimer() { Interval = TimeSpan.FromSeconds(5.0) };
-                dt.Start();
-                dt.Tick += (o, e) => {
-                    dt.Stop();
-                    wnd.Topmost = false;
-                };
-
-                wnd.Closed += (o, e) => dt.Stop();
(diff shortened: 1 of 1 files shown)


50Wliu:
Like you referenced in your commit, this literally removes the only gain that we got in #137.  At least with the taskbar window I could verify that it was still installing (and not crashed, frozen, etc.) by just glancing at the taskbar.  Now I have to minimize who knows how many windows just to get to the GIF (I minimize/maximize windows pretty often), then re-open them.

paulcbetts:
>  Now I have to minimize who knows how many windows just to get to the GIF (I minimize/maximize windows pretty often), then re-open them.

The Window is now Always On Top for the first 5 seconds it exists, so this shouldn't be a problem

50Wliu:
Yes, but Atom takes more than 5 seconds to install.

paulcbetts:
That's fine, after the 5 seconds is up people can put it to the background

50Wliu:
Yes, and that was fine when there was a taskbar icon, because I could bring it up again whenever I wanted to in order to check on it.  Now that's not possible.  Some people may infer that it just...disappeared (and/or crashed) after they get back to whatever they were working on before.

paulcbetts:
Unless we get overwhelming feedback to the contrary during the Slack Windows beta that this is confusing users, this isn't going to change, sorry. I'm getting bugs internally from people who *don't* want  the icon in the taskbar - this is just a matter of personal taste or something.

50Wliu:
I guess so then.

Issue #138, opened 2014-11-25
Couldn't aquire lock, is another instance running?


peters:
After updating to 6.0.1 many of our logs are flooded with these exceptions. Has something changed? Can't spot anything wrong with the code below either. `CheckForUpdates` method is only called from a single place in the application. The update still works though after a couple of retries when it's able to aquire the lock.

/cc @paulcbetts 

```csharp
YggdrasilSetup.CheckForUpdates((version, updateAbsPath) =>
{
                    Info("Successfully updated yggdrasil to new version: {0}. Assembly location is now: {1}", version, updateAbsPath);
                    _updated = true;
                    YggdrasilSetup.OnAppStart(updateAbsPath);
                    YggdrasilAssembly.RestartApplication(updateAbsPath, RestartApplicationReason.Update);
 }).Wait(CancellationToken);
```

```csharp
internal static async Task<bool> CheckForUpdates(Action<Version, string> updateSuccess = null)
        {
            // ReSharper disable once JoinDeclarationAndInitializer
            string updateUrl;

#if INSTALLER
            updateUrl = @"C:\installers\yggdrasil\{0}".Fmt(IntPtr.Size == 4 ? "x86" : "x64");
#elif DEPLOY_PRODUCTION
            updateUrl = "http://software.odinkapital.no/apps/yggdrasil/production/updates";
#elif DEPLOY_STAGING
            updateUrl = "http://software.odinkapital.no/apps/yggdrasil/staging/updates";
#else
            return false;
#endif

#pragma warning disable 162
            if (updateSuccess == null) updateSuccess = (a, b) => { };

            string updateAbsPath = null;
            ReleaseEntry releaseEntry = null;

            using (var updateManager = new UpdateManager(updateUrl, YggdrasilAssembly.AssemblyName, FrameworkVersion.Net45))
            {
                var updateInfo = await updateManager.CheckForUpdate(true);
                if (updateInfo != null && updateInfo.ReleasesToApply.Count > 0)
                {
                    releaseEntry = updateInfo.ReleasesToApply.OrderByDescending(x => x.Version).First();
                    await updateManager.DownloadReleases(updateInfo.ReleasesToApply, i => Logger.Info("DownloadReleases progress: {0}", i));
                    updateAbsPath = await updateManager.ApplyReleases(updateInfo, i => Logger.Info("ApplyReleases progress: {0}", i));
                    updateAbsPath = Path.Combine(updateAbsPath, "yggdrasil.exe");
[...]

PR #189, opened 2015-01-15
SingleGlobalInstance constructor should set the existing IDisposable variable "handle" to overcome locking errors instead of creating a new one and never using it.

theofanis:
I actually don't get why a new handle variable is created in SingleGlobalInstance constructor. When disposing, it's always null, so Interlocked.Exchange(ref handle, null) returns null. As a result if (disp != null) disp.Dispose(); never executes, so the lock file is never deleted. This causes locking errors in subsequent calls to UpdateManager's methods, even after disposing it after each call (wrapping each call in using statements of course).

By setting the existing handle variable instead of creating a new one and never using it, I could overcome my problems as described in #138.

@paulcbetts If my commit causes problems I am not able to see, since I am not really familiar with the project's code, please enlighten me. After this change though, I am able to use Squirrel now :)

### DIFF
--- a/src/Squirrel/Utility.cs
+++ b/src/Squirrel/Utility.cs
@@ -555,7 +555,7 @@ namespace Squirrel
                 throw new Exception("Couldn't acquire lock, is another instance running");
             }
 
-            var handle = Disposable.Create(() => {
+            handle = Disposable.Create(() => {
                 fh.Dispose();
                 File.Delete(path);
             });


paulcbetts:
What a dumb mistake. Thanks @theofanis!

theofanis:
Glad to contribute, squirrel rocks! Can't wait for new release with this fix, my software needs to be deployed!!!

PR #191, opened 2015-01-15
Fix appname install directory

phantomtypist:
Paul, I think I fixed the problem where the directory name of the installed app was not getting set properly if a package Id had a period in it.

Please review :worried:

### DIFF
--- a/src/Squirrel/ReleaseEntry.cs
+++ b/src/Squirrel/ReleaseEntry.cs
@@ -23,6 +23,7 @@ namespace Squirrel
         Version Version { get; }
         string PackageName { get; }
 
+        string GetPackageId(string packageDirectory);
         string GetReleaseNotes(string packageDirectory);
     }
 
@@ -53,9 +54,22 @@ namespace Squirrel
         [IgnoreDataMember]
         public Version Version { get { return Filename.ToVersion(); } }
 
+        [Obsolete("Property is deprecated because it may not correctly retrieve package names (a.k.a. Id's) if you have periods in it.  If you wish to retrieve the package Id, please use the GetPackageId method.")]
         [IgnoreDataMember]
         public string PackageName {
             get { return Filename.Substring(0, Filename.IndexOfAny(new[] { '-', '.' })); }
+        }
+
+        public string GetPackageId(string packageDirectory)
+        {
+            var zp = new ZipPackage(Path.Combine(packageDirectory, Filename));
+
+            if (String.IsNullOrWhiteSpace(zp.Id))
+            {
+                throw new Exception(String.Format("Invalid 'Id' value in nuspec file at '{0}'", Path.Combine(packageDirectory, Filename)));
+            }
+
+            return zp.Id;
         }
 
         public string GetReleaseNotes(string packageDirectory)
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -162,7 +162,7 @@ namespace Squirrel.Update
             }
 
             var ourAppName = ReleaseEntry.ParseReleaseFile(File.ReadAllText(releasesPath, Encoding.UTF8))
-                .First().PackageName;
+                .First().GetPackageId(sourceDirectory);
 
             using (var mgr = new UpdateManager(sourceDirectory, ourAppName, FrameworkVersion.Net45)) {
                 Directory.CreateDirectory(mgr.RootAppDirectory);
--- a/test/ReleaseEntryTests.cs
+++ b/test/ReleaseEntryTests.cs
@@ -69,6 +69,18 @@ namespace Squirrel.Tests.Core
             Assert.Equal(expectedMinor, fixture.Version.Minor);
         }
 
+        [Theory]
+        [InlineData("Squirrel.Core.1.0.0.0.nupkg", "NSync.Core")]
+        [InlineData("Squirrel.Core.NoDependencies.1.0.0.0", "NSync.Core")]
+        [InlineData("ProjectWithContent.1.0.0.0-beta.nupkg", "ProjectWithContent")]
+        public void ParsePackageIdFromReleaseEntryTest(string packageFileName, string packageId)
+        {
+            var path = IntegrationTestHelper.GetPath("fixtures", packageFileName);
+            var fixture = ReleaseEntry.GenerateFromFile(path);
+
+            Assert.Equal(packageId, fixture.GetPackageId(IntegrationTestHelper.GetPath("fixtures")));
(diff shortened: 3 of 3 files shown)


src/Squirrel/ReleaseEntry.cs
```diff
@@ -53,11 +54,24 @@ protected ReleaseEntry(string sha1, string filename, long filesize, bool isDelta
         [IgnoreDataMember]
         public Version Version { get { return Filename.ToVersion(); } }
 
+        [Obsolete("Property is deprecated because it may not correctly retrieve package names (a.k.a. Id's) if you have periods in it.  If you wish to retrieve the package Id, please use the GetPackageId method.")]
         [IgnoreDataMember]
         public string PackageName {
             get { return Filename.Substring(0, Filename.IndexOfAny(new[] { '-', '.' })); }
```
paulcbetts: Why don't you just fix this method to correctly handle IDs with '.'s in it?

phantomtypist: I was a little worried about breaking functionality for people that might be using it and relying on it to work the way it was working.

Another thought was I didn't want to make any more assumptions about parsing a filename to find the package Id/Name... too many unknowns to worry about.  I figured it was just best to go to the source of where it is inside the NuGet package metadata...  kinda like how you are getting the release notes.

My last thought was that "package Id" was a better fit than "package name".  Now, I know that in the description of the "package Id" in the Nuget nuspec, they refer to the Id as the name of the package.  However, I've run into people in the past that have gotten confused and associated a "package name" with the actual title of the package.

Let me know if you want me to make some modifications.

paulcbetts: > I was a little worried about breaking functionality for people that might be using that property and relying on it to work the way it was working.

This is a good impulse but this API isn't *actually* intended for anyone to be used, and at the moment, it is returning incorrect results (i.e. nobody can be relying on it returning an incorrect package name because it's not returning anything useful for those people)

> Another thought was I didn't want to make any more assumptions about parsing a filename to find the package Id/Name... too many unknowns to worry about.

This is a fundamental part of Release Entry, that the name *must be* of the format "PackageName-Version-[delta/full].nupkg". This is because when we read the RELEASES file, we don't have access to the NuGet package, so we can't read the metadata.

phantomtypist: Alright, let me take another crack at it.

phantomtypist: My one thought was the format "PackageName-Version-[delta/full].nupkg" could be hard to parse. You could have package names with periods and hyphens... and also at the same time the version could be semantic with hyphens too (e.g. 1.1.0.0-beta.)

Would you be opposed to me using a regular expression?

paulcbetts: > You could have package names with periods and hyphens

Yeah, probably easiest to start at the back and parse towards the front. 

> and also at the same time the version could be semantic with hyphens too (e.g. 1.1.0.0-beta.)

Prerelease tags aren't supported, ignore that case

> Would you be opposed to me using a regular expression?

Not at all :)

phantomtypist: > Prerelease tags aren't supported, ignore that case

Can you help me understand why prerelease tags aren't supported?

paulcbetts: Because we use the "slot" for prerelease tags to tag as "delta" or "full" package

phantomtypist: > Because we use the "slot" for prerelease tags to tag as "delta" or "full" package

I get it now.  I was in the mindset thinking the "delta" and "full" tags were extra stuff outside of the version scope.

phantomtypist: Since prerelease tags aren't supported, is there an actual way/strategy to release prerelease applications (i.e. beta)?

paulcbetts: Use a different release channel - Squirrel was built s.t. you can easily set up Channels a-la Chrome / Firefox, where different users get different sets of updates, just by changing the folder / URL you update from

phantomtypist:
@paulcbetts FYI I'm done making changes.  Ready for review.

src/Squirrel/ReleaseEntry.cs
```diff
@@ -53,9 +53,19 @@ protected ReleaseEntry(string sha1, string filename, long filesize, bool isDelta
         [IgnoreDataMember]
         public Version Version { get { return Filename.ToVersion(); } }
 
+        static readonly Regex packageRegex = new Regex(@"^(?<packageName>.*)-(?<version>\d+\.\d+\.\d+\.\d+)-(?<releaseType>full|delta).nupkg$");
         [IgnoreDataMember]
         public string PackageName {
-            get { return Filename.Substring(0, Filename.IndexOfAny(new[] { '-', '.' })); }
+            get
+            {
```
paulcbetts: Move this bracket on the previous line

src/Squirrel/ReleaseEntry.cs
```diff
@@ -53,9 +53,19 @@ protected ReleaseEntry(string sha1, string filename, long filesize, bool isDelta
...
+                var match = packageRegex.Match(Filename);
+
+                if (!match.Success) {
```
paulcbetts: Can you move this validation into the place where we provide this information (i.e. the ctor?) Getters that throw are evil

phantomtypist: It's gonna break a lot of the tests because the InlineData for the test has invalid package names in certain ones.

Do you mind if I change those as well?

phantomtypist: Every filename in the release entry will always contains "-delta|-full" at the end before the ".nupkg" extension?

paulcbetts: Yes, that's guaranteed

JakeGinnivan:
Would love this to be merged, would have saved me a bit of time.

Pjanssen:
Is there any chance of this PR being merged? 
I'm running into issues where I have multiple applications with a package ID in this format: `CompanyName.Product.SubProduct`. All of these subproducts are getting installed into the same directory `CompanyName`.

phantomtypist:
@Pjanssen I'm busy with life at the moment, but I'll see if I can find time in the near future, unless you want to finish it for me?

Pjanssen:
@phantomtypist Could you describe what is left to be done?

PR #201, opened 2015-01-22
Specify Shortcut Locations on Update.exe

CharlieHess:
New argument that gives users control of the shortcuts that Update.exe manages.
Usage: 
`"l=|shortcut-locations=", "Comma-separated string of shortcut locations, e.g. 'Desktop,StartMenu'"`

### TODO:
- [x] Test this thoroughly
- [ ] Code review

### DIFF
--- a/src/Squirrel/IUpdateManager.cs
+++ b/src/Squirrel/IUpdateManager.cs
@@ -13,6 +13,7 @@ namespace Squirrel
 {
     [Flags]
     public enum ShortcutLocation {
+        None = 0,
         StartMenu = 1 << 0,
         Desktop = 1 << 1,
         Startup = 1 << 2,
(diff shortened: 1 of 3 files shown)


src/Squirrel/IUpdateManager.cs
```diff
@@ -13,6 +13,7 @@ namespace Squirrel
 {
     [Flags]
     public enum ShortcutLocation {
+        None = 0,
```
paulcbetts: Should we just have a default list? What is the intent of using None?

CharlieHess: Just best practices for `Flags` enums. There is a default list specified if the user passes null or an empty string. Either way works for me, though.

PR #205, opened 2015-01-27
Sync from standard http

### LINKED ISSUE #194: Proposal: Better workflow for creating delta packages
One of the nicest things about Squirrel is the fact that it can trivially create and apply delta packages.

At the moment, this is tricky to do from a CI environment. Our build works as follows:

1. Use TC to perform a clean checkout of the our product repo in GitHub
2. Build our product (`c#` based)
3. Test our product
4. Build an installer using Squirrel

Squirrel is building an installer from freshly compiled artifacts and doesn't have access to any previous `nupkg`s as we are working in a newly created clean TC checkout.

The current workflow means that I need to add a step between 3 and 4 which downloads the most recently uploaded `nupkg` and `RELEASES` from our S3 account and drops it in the `Releases` directory in TC.

My proposal is that an additional parameter be added to `--realeaseify` that points to an existing RELEASES url. `Releaseify` woulud then parse that, download the latest `nupkg` and therefore always be in a position to create both the `-full` and `-delta`

Does this seem like a sensible addition, or is there something about my workflow that means we are using Squirrel incorrectly?

kevfromireland:
Resolves #194

1. Reuses the `-u` arg and the presence of the GitHub token is what is used to choose between syncing with GitHub or treating the url as a standard http location

2. Didn't implement retry although can later if needed.

3. Tried to make as few changes to the existing code as possible.

4. Skips an existing `nupkg` if it exists on disk. Didn't bother with SHA1 checking

### DIFF
--- a/src/SyncGitHubReleases/Program.cs
+++ b/src/SyncGitHubReleases/Program.cs
@@ -59,27 +59,28 @@ namespace SyncGitHubReleases
                     return -1;
                 }
 
+                var releasesDirectoryInfo = new DirectoryInfo(releaseDir ?? Path.Combine(".", "Releases"));
+                
+                if (!releasesDirectoryInfo.Exists) 
+                    releasesDirectoryInfo.Create();
+
                 if (token == null)
                 {
                     Console.WriteLine("No GitHub token specified so assuming URL points to existing RELEASES file");
-                    await SyncFromStandardHttp(repoUrl, releaseDir ?? Path.Combine(".", "Releases"));
-
+                    await SyncFromStandardHttp(new Uri(repoUrl), releasesDirectoryInfo);
                 }
                 else
                 {
                     Console.WriteLine("GitHub token was specified, so will create Releases directory from repository");
-                    await SyncFromGitHub(repoUrl, token, releaseDir ?? Path.Combine(".", "Releases"));
+                    await SyncFromGitHub(repoUrl, token, releasesDirectoryInfo);
                 }
             }
 
             return 0;
         }
 
-        private async Task SyncFromGitHub(string repoUrl, string token, string releaseDir)
+        private async Task SyncFromGitHub(string repoUrl, string token, DirectoryInfo releaseDirectoryInfo)
         {
-            var releaseDirectoryInfo = new DirectoryInfo(releaseDir);
-            if (!releaseDirectoryInfo.Exists) releaseDirectoryInfo.Create();
-
             var repoUri = new Uri(repoUrl);
             var userAgent = new ProductHeaderValue("SyncGitHubReleases", Assembly.GetExecutingAssembly().GetName().Version.ToString());
             var client = new GitHubClient(userAgent, repoUri)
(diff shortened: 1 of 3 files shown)


src/SyncGitHubReleases/Program.cs
```diff
@@ -62,65 +62,87 @@ public static int Main(string[] args)
...
+            {
+                Credentials = new Credentials(token)
+            };
+
+            var nwo = nwoFromRepoUrl(repoUrl);
+            var releases = (await client.Release.GetAll(nwo.Item1, nwo.Item2))
+                .OrderByDescending(x => x.PublishedAt)
+                .Take(2);
+
+            await releases.ForEachAsync(async release =>
```
paulcbetts: Yo why'd you change the bracket layout everywhereeeeee

src/SyncGitHubReleases/StandardHttpSyncer.cs
```diff
@@ -0,0 +1,74 @@
...
+            var releasesIndex = await DownloadReleasesIndex(m_Uri);
+
+            File.WriteAllText(Path.Combine(releasesDir.FullName, "RELEASES"), releasesIndex);
+
+            var releasesToDownload = ReleaseEntry.ParseReleaseFile(releasesIndex)
+                .Where(x => !x.IsDelta)
+                .OrderByDescending(x => x.Version)
+                .Take(2)
+                .Select(x => new
+                             {
```
paulcbetts: How can R# users tolerate stuff like this?

src/SyncGitHubReleases/StandardHttpSyncer.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            if (File.Exists(localPath))
+            {
+                Console.WriteLine("Skipping this release as existing file found at: {0}", localPath);
+                return;
+            }
+
+            Console.WriteLine("Downloading release from {0}", remoteUrl);
+
+            using (HttpClient client = new HttpClient())
```
paulcbetts: This might be easier using `WebClient.DownloadFileTaskAsync`

src/SyncGitHubReleases/Program.cs
```diff
@@ -48,13 +48,13 @@ public static int Main(string[] args)
...
                     { "h|?|help", "Display Help and exit", _ => {} },
                     { "r=|releaseDir=", "Path to a release directory to download to", v => releaseDir = v},
-                    { "u=|repoUrl=", "The URL to the repository root page", v => repoUrl = v},
+                    { "u=|repoUrl=", "When pointing to GitHub, use the URL to the repository root page, else point to an existing RELEASES URL", v => repoUrl = v},
                     { "t=|token=", "The OAuth token to use as login credentials", v => token = v},
                 };
 
                 opts.Parse(args);
 
-                if (token == null || repoUrl == null || repoUrl.StartsWith("http", true, CultureInfo.InvariantCulture) == false) {
```
paulcbetts: Instead of checking for the token, can you just try to treat the URL as a GitHub repo, and if it fails, try to treat it as a RELEASES folder? I actually always set the token via an environment variable in my builds

paulcbetts:
This is great! Just a few things to clean up and we're good to go

src/SyncGitHubReleases/Program.cs
kevfromireland: Mistakenly copy and pasted in VS and it 'fixed' it up for me. Copy and pasted back from the original file using notepad.

src/SyncGitHubReleases/StandardHttpSyncer.cs
kevfromireland: Bracket layout? I have zero preference on this and comfortable bouncing back and forth. Do you have a style you would prefer to use? I tend to write what I see as most idiomatic for c# taking my cue from the BCL authors, but literally don't care. Reformat away!

src/SyncGitHubReleases/StandardHttpSyncer.cs
kevfromireland: Happy with either. Took my cue from Program.cs which uses HttpClient to download the Github release. 

Want me to change it to use WebClient, or leave it consistent with Program.cs?

src/SyncGitHubReleases/Program.cs
kevfromireland: This feels pretty dirty. As Squirrel probably doesn't have many users yet, should we do the right thing now and add an explicit `-mode` switch or something?

src/SyncGitHubReleases/StandardHttpSyncer.cs
paulcbetts: If the current code here works it's probably fine, more of a Tip™

src/SyncGitHubReleases/Program.cs
paulcbetts: Nope, feels great to me :)

kevfromireland: So do a random call using `GitHubClient` and then catch exceptions?

paulcbetts: No, just try to run the `SyncFromGitHub` method and if it doesn't work, try `Sync` then if it really doesn't work, blow up

kevfromireland: I will do this if you really, really, really want me to but I hate it.

1. Syncing from a standard location is now part of the happy path of this program, no need to treat it as anything other than happy path

2. You are leaking your auth token now to a random website. Hopefully under your control so you don't care but just feels so wrong (Maybe putting your auth token over `http`?

3. Makes this more brittle in the future as SyncFromGitHub becomes harder to change. Side effects in its init can happy now even if you *know* you never intended them.

I can do it, but I just don't see the upside.

paulcbetts:
@kevfromireland Merged with some fixups, thanks @kevfromireland!

PR #214, opened 2015-02-02
Enable more precise download progress

sickboy:
Gives a bit more interactive feeling to the user especially when he is farther behind or larger updates got released.


BarryThePenguin:
Is there anyway to include this in the Update/AnimatedGifWindow.cs as the progress indicator too? Then show/hide the progress indicator based on a configuration option? :wink:

paulcbetts:
@BarryThePenguin This won't help initial install, since we're not downloading anything, the biggest consumer of time is NuGet extracting the archives, which I _guess_ we could get progress on but I'm not super excited about that. I *would* take a PR to add progress to the Win7 Taskbar (it wouldn't be optional)

BarryThePenguin:
Ah yes, of course

sickboy:
@paulcbetts Win7 progress-bar for Setup.exe, or for when running updates in your client?
The latter sounds like a bad idea (TM), due to interference with the application, so I'm guessing the former ;-)

paulcbetts:
@sickboy Only for the former - it's on the app developer if they want it for updates

sickboy:
@paulcbetts something like this then?
```
                var taskbarItemInfo = new TaskbarItemInfo {
                    ProgressState = TaskbarItemProgressState.Normal,
                };
                var app = new Application();
                var window = new Window() {
                    TaskbarItemInfo = taskbarItemInfo,
                    WindowState = WindowState.Minimized,
                    Width = 0,
                    Height = 0,
                    WindowStyle = WindowStyle.None
                };
                app.Startup += async (sender, args) => {
                    await mgr.FullInstall(silentInstall, p => app.Dispatcher.Invoke(() => taskbarItemInfo.ProgressValue = p / 100.0));

                    await this.ErrorIfThrows(() =>
                        mgr.CreateUninstallerRegistryEntry(),
                        "Failed to create uninstaller registry entry");
                    app.Shutdown();
                };
                app.Run(window);
```
I wonder if it's possible without a WPF Window ;-)

PS are you interested in this pull-request as-is, need some changes first, or not at all interested?

paulcbetts:
> @paulcbetts are you interested in this pull-request as-is, need some changes first, or not at all interested?

Yeah, definitely - at the time it needed a merge fixup. I'll have a look today

> Regarding the taskbar progress info, something like this then?

We've already got a WPF window in AnimatedGIFWindow for Setup. 

>     await mgr.FullInstall(silentInstall, p => app.Dispatcher.Invoke(() => window.TaskbarItemInfo.ProgressValue = p / 100.0));

^^^ Always use BeginInvoke, but you've got the right idea :)

sickboy:
> Yeah, definitely - at the time it needed a merge fixup. I'll have a look today
Awesome!

> Always use BeginInvoke, but you've got the right idea :)
Sure thing!

I've opened pull request #216 in the meantime, so we can better track it there. The code is what I posted here.
Will look into the AnimatedGIFWindow.

BarryThePenguin:
:tada: :balloon:

PR #215, opened 2015-02-02
Webclients aughta be disposed

sickboy:
Can haz dispose? http://stackoverflow.com/questions/1832992/calling-dispose-vs-when-an-object-goes-out-scope-method-finishes


paulcbetts:
Into it

paulcbetts:
Thanks @sickboy!

PR #216, opened 2015-02-04
Add install progress

sickboy:
Experimental at best.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -120,11 +120,12 @@ namespace Squirrel.Update
 
                 switch (updateAction) {
                 case UpdateAction.Install:
+                    var progressSource = new ProgressSource();
                     if (!silentInstall) { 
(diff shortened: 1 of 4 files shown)


src/Update/Program.cs
```diff
@@ -181,11 +184,25 @@ int main(string[] args)
                 this.ErrorIfThrows(() => File.Copy(Assembly.GetExecutingAssembly().Location, updateTarget, true),
                     "Failed to copy Update.exe to " + updateTarget);
 
-                await mgr.FullInstall(silentInstall);
+                var app = new Application();
```
paulcbetts: Why are we creating a dummy window? We've already got a Window in AnimatedGifWindow

sickboy:
Using the existing window now, probably could refactor a little more to move the app/window logic out of the Install method.

sickboy:
Decoupled and cleaned up now. I miss RX ;)

paulcbetts:
Into it. Thanks @sickboy!@

sickboy:
Awesome, thanks @paulcbetts!

PR #224, opened 2015-02-10
Do not send null token in Authorization Header

BarryThePenguin:
Otherwise we get 401 not authorized as a result...


paulcbetts:
Thanks!

PR #229, opened 2015-02-13
Use GitHub url for UpdateManager

### LINKED ISSUE #223: UpdateManager CheckForUpdate from GitHub
Would it be worth including something similar to the GitHub implementation in SyncReleases in UpdateManager/CheckForUpdate methods?

BarryThePenguin:
closes #223

### DIFF
--- a/src/Squirrel/Squirrel.csproj
+++ b/src/Squirrel/Squirrel.csproj
@@ -78,6 +78,8 @@
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Drawing" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Net.Http.Formatting, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL" />
     <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.Xml.Linq" />
     <Reference Include="Microsoft.CSharp" />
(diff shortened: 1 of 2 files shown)


paulcbetts:
Can you just use HttpClient instead of including Octokit? I'm trying to limit the dependencies of Squirrel.dll as much as possible

BarryThePenguin:
I thought this functionality would fit better in squirrel because of the dependency. But I didn't think of httpclient.

Is the static class factory suitable? Or should I create a new factory class? Updatemanager is already quite large...

BarryThePenguin:
Now using HttpClient and a custom Release object with relevant properties.

BarryThePenguin:
A note on [HttpClient](http://stackoverflow.com/questions/15705092/do-httpclient-and-httpclienthandler-have-to-be-disposed)

paulcbetts:
I'll have a look at this soon

BarryThePenguin:
Heh... I just realised this doesn't support updates multiple and deltas...

BarryThePenguin:
Although, during the update process, it did download both the new -full and -delta nupkg

BarryThePenguin:
Any thoughts? Otherwise I'll just maintain this in the app

paulcbetts:
Remember when I said I would look at this soon then _totally_ spaced on it? This time I'll actually look into it :)

BarryThePenguin:
Sure, just looking for feedback. Would be good to know if this is a Good Idea™ or not ;)

Cyral:
This is a great idea, I was about to suggest it myself because I needed it and then found this!

paulcbetts:
:+1:

ricardopolo:
So how to use it? 
It will download relases from Github Releases or what?

BarryThePenguin:
Should this be documented? Where should the documentation be? Comments in code or a wiki page or similar?

paulcbetts:
> Should this be documented? Where should the documentation be? Comments in code or a wiki page or similar?

Throw this up on a Wiki page if you would!

BarryThePenguin:
I added it to the Client Implementation docs #391

PR #232, opened 2015-02-15
Master

MartinRuden:
Avoid crashing when the user doesn't supply any shortcutLocations.


paulcbetts:
I believe something has gone pear-shaped with Git here, can you run:

```sh
git remote add upstream https://github.com/squirrel/squirrel.windows
git fetch upstream
git rebase upstream/master
```

PR #247, opened 2015-02-27
Fix infinite shortcut issues

paulcbetts:
This PR fixes a bunch of issues around shortcuts and um other stuff

* Kill Squirrel Event processes that get cancelled
* We set the AppID which should convince pins to not get dupes showing up when people pin an old version then update to a new version
* Instead of creating shortcuts to the EXE itself, we instead create a shortcut to Update.exe which then always runs the latest version of the EXE
* Instead of always recreating shortcuts, we'll leave them alone if they're already pointing to what we want

## TODO:

- [x] Make sure this actually works

### DIFF
--- a/src/Squirrel/ShellFile.cs
+++ b/src/Squirrel/ShellFile.cs
@@ -8,17 +8,11 @@ using System.Windows.Forms;
 
 namespace Squirrel
 {
-    #region ShellLink Object
-
     /// <summary>
     /// Summary description for ShellLink.
     /// </summary>
     public class ShellLink : IDisposable
     {
-        #region ComInterop for IShellLink
-
-        #region IPersist Interface
-
         [ComImport()]
         [Guid("0000010C-0000-0000-C000-000000000046")]
         [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
@@ -28,10 +22,6 @@ namespace Squirrel
             //[helpstring("Returns the class identifier for the component object")]
             void GetClassID(out Guid pClassID);
         }
-
-        #endregion
-
-        #region IPersistFile Interface
 
         [ComImport()]
         [Guid("0000010B-0000-0000-C000-000000000046")]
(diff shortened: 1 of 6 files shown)


src/Squirrel/UpdateManager.ApplyReleases.cs
```diff
@@ -149,17 +152,35 @@ public void CreateShortcutsForExecutable(string exeName, ShortcutLocation locati
...
+                        sl = new ShellLink {
+                            Target = updateExe,
                             IconPath = exePath,
                             IconIndex = 0,
                             WorkingDirectory = Path.GetDirectoryName(exePath),
                             Description = zf.Description,
+                            Arguments = "--processStart " + exeName,
                         };
 
+                        sl.SetAppUserModelId(String.Format("com.squirrel.{0}.{1}", zf.Id, exeName.Replace(".exe", "")));
```
BrandonLive: FYI, the convention for AppUserModelIds is:
CompanyName.ProductName.SubProduct.VersionInformation

Since Squirrel appears to be your deployment library, I don't think it makes sense to use that here, but rather just use a full value provided by the app.

paulcbetts: The problem is that we don't really have that value in a way I can guarantee to be safe

src/Squirrel/ShellFile.cs
```diff
@@ -64,9 +54,46 @@ interface IPersistFile
                 [MarshalAs(UnmanagedType.LPWStr)] out string ppszFileName);
         }
 
-        #endregion
+        [StructLayout(LayoutKind.Sequential)]
+        public struct PropVariant
+        {
+            public short variantType;
+            public short Reserved1, Reserved2, Reserved3;
```
BrandonLive: These should be "ushort"

src/Squirrel/ShellFile.cs
```diff
@@ -863,6 +852,12 @@ public Keys HotKey
             }
         }
 
+        public void SetAppUserModelId(string appId)
+        {
+            var propStore = (IPropertyStore)linkW;
+            propStore.SetValue(PROPERTYKEY.PKEY_AppUserModel_ID, appId);
```
BrandonLive: SetValue takes a PropVariant, not a string. You'll need to create and init a PropVariant to set this. Normally you'd use InitPropVariantFromString for this, but that's an inline function in propvar.h. I'd look at what the API Code Pack version does as it was tested and should have utilities for making this easy.

src/Squirrel/ShellFile.cs
```diff
@@ -863,6 +852,12 @@ public Keys HotKey
             }
         }
 
+        public void SetAppUserModelId(string appId)
+        {
+            var propStore = (IPropertyStore)linkW;
```
BrandonLive: Does C# know to do a QI for a cast like this? Or do you need Marhsal.QueryInterface or something like that?

paulcbetts: Seems to, yeah

src/Squirrel/ShellFile.cs
paulcbetts: That's what I thought too, but there are C++ code samples on The Old New Thing that pass constant strings as the 2nd argument. It seems to work.

src/Squirrel/UpdateManager.ApplyReleases.cs
BrandonLive: Hmm. That's unfortunate. There's no package name or anything that would be appropriate? Or no metadata field you could add for this? (and only apply to the shortcut if they specified it?)

paulcbetts: `id` is the only NuGet package variable that has restrictions on it (i.e. no weird characters or spaces), the rest of them are designed to be human readable

src/Squirrel/ShellFile.cs
BrandonLive: Could you link me to one of those? That doesn't sound right at all...

paulcbetts: http://blogs.msdn.com/b/oldnewthing/archive/2012/08/17/10340743.aspx

BrandonLive: That is using the IPropertyStore_SetString helper. You are not.

BrandonLive: (which oddly I think might be an internal-only helper. whoops)

BrandonLive: Oh, he links it to another post where he shared the implementation:
http://blogs.msdn.com/b/oldnewthing/archive/2011/06/01/10170113.aspx

PR #249, opened 2015-02-28
Log URLs correctly in FileDownloader

willdean:
Expressions of the form `"Downloading file: " + failedUrl ?? url` evaluate the `+` before the `??`, which means that they never actually print `url`, just `failedUrl`.  This change fixes that.


paulcbetts:
Thanks!

Issue #231, opened 2015-02-15
Temporary Files Overload


karai17:
Two days ago I had an issue with Atom that required me to reinstall it. Today I am doing a routine clean out of my temp files and I have 42,802 empty .tmp files dating to the date and time I installed Atom. While I am not 100% certain it was Atom/Squirrel, it is high on my list of suspects as I have not installed or updated any other software recently.

The files are all named in the same format: `tmpA7E3.tmp`, where `A7E3` are hex values. Every single file is 0kb so I suspect there may be an issue with deleting these files, not only their contents.

karai17:
To add to this, there was also a nuget folder in there with 77,182 files totalling 1GB with the same date.

paulcbetts:
I recommend that you delete them

karai17:
I already did, but shouldn't Squirrel clean up after itself?

paulcbetts:
Probably

shiftkey:
Closing this out because #250 was merged, and it's not clear to me if there's any extra work that should be done in here.

Please open a fresh issue with more details if this is seen with a more recent version of Squirrel.

PR #250, opened 2015-02-28
Added deletion guard for temporary file

willdean:
After reading issue #231, I had a quick look for references to `Path.GetTempFileName()` to look for leaks.  The first one I found was in `DeltaPackageBuilder.applyDiffToFile`.   This PR is a proposed fix for this - it removes an internal try/catch block (with a File.Delete-and-rethrow), and replaces it with an overall try/finally, which closes at least two routes by which temporary files might be leaked.

It looks like a huge patch because of the changing indent, it's only a couple of lines, really.

I've not done any testing on this, and I've not reviewed any of the other uses of `Path.GetTempFileName` for the same sort of issues.    I'm happy to go through them all and and do the same sort of thing if you'd like.

Another approach I've used in other projects is to make a simple wrapper for temporary files, which is both IDisposable AND implements a finalizer to delete the file if you forget to Dispose it - I'd be happy to do something like that instead if you preferred.

### DIFF
--- a/src/Squirrel/DeltaPackage.cs
+++ b/src/Squirrel/DeltaPackage.cs
@@ -187,38 +187,40 @@ namespace Squirrel
             var finalTarget = Path.Combine(workingDirectory, Regex.Replace(relativeFilePath, @".diff$", ""));
 
             var tempTargetFile = Path.GetTempFileName();
+            try
+            {
+                // NB: Zero-length diffs indicate the file hasn't actually changed
+                if (new FileInfo(inputFile).Length == 0) {
+                    this.Log().Info("{0} exists unchanged, skipping", relativeFilePath);
+                    return;
+                }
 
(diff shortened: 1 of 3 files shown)


paulcbetts:
The temp file leak is almost certainly in NuGet's OptimizedZipPackage file, but this is still good.

willdean:
Paul, sorry, with some horror I see a bunch of stuff in that PR which wasn't there when I created it (I am clearly miles off the pace when it comes to understanding how/why GH updates existing PRs with changes made between their creation and when they're merged).  You might want to back this out, as it had some stuff in there you probably don't want (e.g. fd44e7e, which might have been the subject of a future, unrelated, PR).  Sorry about this.

paulcbetts:
@willdean Nah, I saw that, I'm into it.

Issue #252, opened 2015-03-02
.NET4.5 installation needs, but doesn't get, reboot on clean Win7 install


willdean:
I'm running a Squirrel-based installer on a completely clean Win7 x64 machine (in a VM, so it's easy to revert to this state repeatedly)

The Squirrel installer correctly detects that it needs to install .Net 4.5.  It downloads it, runs the installer which seems to run fine, then just silently stops at the end.   Nothing else happens, the app is not installed.

In this situation, if you then try and run the installer .exe again, it fails with 0xc000007b, offers to show the squirrel setup log (which doesn't work/isn't there), and then exits.  

Windows then shows the "it looks like an installer didn't run correctly, would you like to try again with the recommended settings?" message.

If you reboot the machine, the .NET installation finishes, and you can now re-run the installer, which runs to completion OK.

I think the problem here is that the `/norestart` option which setup.exe passes to the framework installer actually means "tell me if we need to restart but don't do it yourself", not "do your magic without ever needing a restart".

From here:  https://msdn.microsoft.com/en-us/library/ee942965%28v=vs.110%29.aspx I deduce that Squirrel needs to handle return codes of 1641 and 3010 as being success-requiring-a-reboot, not as failure.   Given that the rest of the install can't finish without that reboot, there then either needs to be some runonce-type stuff to re-run the installer after restart (sigh), or instructions to the user to re-run themselves after reboot (yuk, yuk).  

I'm happy to provide a PR to improve this in some way.  Suggestions?

I guess *clean* Win 7 machines are a pretty rare thing nowadays...

willdean:
Doing the runonce thing didn't turn out to be too hard, so I've stuck in a quick PR - I've tried this on a Win7 clean VM and it can now install successfully.

PR #253, opened 2015-03-02
Reboot if necessary after .NET framework installation

willdean:
See #252 - this PR checks for the .NET framework installer return codes which indicate that a restart is required (without which the rest of the Squirrel install will fail).  

It then sets up a runonce entry to continue the installation after reboot, and does a reboot.   The installer will re-run after a reboot, and proceed as normal, because the framework is now available.

### DIFF
--- a/src/Setup/FxHelper.cpp
+++ b/src/Setup/FxHelper.cpp
@@ -253,7 +253,12 @@ out:
 // Deal with the aftermath of the framework installer telling us that we need to reboot
 bool CFxHelper::HandleRebootRequirement(bool isQuiet)
 {
-	if (!isQuiet) {
+	if (isQuiet) {
+		// Don't silently reboot - just error-out
+		fprintf_s(stderr, "A reboot is required following .NET installation - reboot then run installer again.\n");
+		return false;
+	}
+	else {
 		CTaskDialog dlg;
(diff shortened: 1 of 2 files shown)


willdean:
Possible issue with this is that if `isQuiet` is set, then this will reboot without asking permission.  Not sure if that's a good thing or not.

paulcbetts:
> Possible issue with this is that if isQuiet is set, then this will reboot without asking permission. Not sure if that's a good thing or not.

That's a bad thing, we should print a "We have to reboot sry" message to stderr, then exit with a non-zero error code

paulcbetts:
This is a great idea, thanks for the PR!

willdean:
OK - I've made that change - PR's updated.  Just did a fprintf_s to stderr - not really sure if that's better or worse than OutputDebugString for a non-console app.

paulcbetts:
I merged this with a few fixups, thanks!

willdean:
I don't think 'well never get here because reboot' is right - the ExitWindowsEx call is async, and returns normally, even though a reboot has been scheduled - so I still think you want to treat this as some kind of blocking failure - perhaps using the same S_FALSE mechanism as for the message box?

paulcbetts:
@willdean Fixed in 9e41fc3

PR #256, opened 2015-03-03
Fix #251

### LINKED ISSUE #251: Unable to update when RELEASES file contains multiple entries with the same version
When Squirrel releasifies the same version multiple times, the RELEASES file contains multiple entries with the same version. When updating the application, it crashes because at some point it only expected a single release entry.
A fix could be to remove existing entries with the same version from the RELEASES file when releasifying the app.
Here is the important part of the stack trace:

    System.InvalidOperationException: Sequence contains more than one matching element
       at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
       at Squirrel.Utility.FindCurrentVersion(IEnumerable`1 localReleases)
       at Squirrel.UpdateManager.CheckForUpdateImpl.determineUpdateInfo(IEnumerable`1 localReleases, IEnumerable`1 remoteReleases, Boolean ignoreDeltaUpdates)
       at Squirrel.UpdateManager.CheckForUpdateImpl.<CheckForUpdate>d__28.MoveNext()

eggapauli:
I also changed the type of `previousReleases` to be a list to avoid multiple enumerations.

I'm not sure if I also should create some tests. Currently there is no test project for the "Update" project. All tests seem to test functionality on the "customer"-side.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -339,9 +339,9 @@ namespace Squirrel.Update
             var processed = new List<string>();
 
             var releaseFilePath = Path.Combine(di.FullName, "RELEASES");
-            var previousReleases = Enumerable.Empty<ReleaseEntry>();
+            var previousReleases = new List<ReleaseEntry>();
             if (File.Exists(releaseFilePath)) {
-                previousReleases = ReleaseEntry.ParseReleaseFile(File.ReadAllText(releaseFilePath, Encoding.UTF8));
+                previousReleases.AddRange(ReleaseEntry.ParseReleaseFile(File.ReadAllText(releaseFilePath, Encoding.UTF8)));
             }
 
             foreach (var file in toProcess) {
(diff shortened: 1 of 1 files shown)


willdean:
@eggapauli I would be happy to help you create some more tests if Paul agrees they're worthwhile - even a lot of the existing Squirrel tests are disabled/failing.  

I think a problem might be that there could be a load more abstraction required to make more things testable, and that may be  unpopular.

eggapauli:
@willdean I agree that tasks like releasifying should not be in Update.exe. But I would prefer not to include it in Squirrel.dll, but create a separate project for it. Then we could use this for example on a build server to releasify apps without having to start a separate process (i.e. Squirrel.exe). But that's not the topic of this PR anyway.

paulcbetts:
Thanks! :+1:

PR #260, opened 2015-03-03
Minor API doc tweak

willdean:
I *think* this is right - I can't find any route where it could return null, and it never has to me.


paulcbetts:
:+1:

Issue #239, opened 2015-02-21
Squirrel fails to delete some folders


50Wliu:
So I was mucking around in `SquirrelSetup.log` today after wondering why Atom was taking so long to update (realized it was because I was a few versions behind and Squirrel had to create deltas for every newer version).

Anyway, I noticed that it threw a few errors near the end while deleting the old version of Atom:
First exception:

```
Utility: DeleteDirectory: could not delete - C:\Users\<User>\AppData\Local\atom\app-0.174.0\resources\app\node_modules\language-todo: System.IO.IOException: The directory is not empty.

   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound)
   at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
   at Squirrel.Utility.<DeleteDirectory>d__36.MoveNext()
```

Second exception block:

```
Utility: DeleteDirectory: could not delete - C:\Users\<User>\AppData\Local\atom\app-0.174.0\resources\app\node_modules: System.IO.IOException: The directory is not empty.

   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound)
   at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
   at Squirrel.Utility.<DeleteDirectory>d__36.MoveNext()
Utility: DeleteDirectory: could not delete - C:\Users\<User>\AppData\Local\atom\app-0.174.0\resources\app: System.IO.IOException: The directory is not empty.

   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound)
   at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
   at Squirrel.Utility.<DeleteDirectory>d__36.MoveNext()
Utility: DeleteDirectory: could not delete - C:\Users\<User>\AppData\Local\atom\app-0.174.0\resources: System.IO.IOException: The directory is not empty.

[...]

PR #263, opened 2015-03-04
Update.exe - Add timestamp to log messages, and log unhandled exceptions

willdean:
We are really struggling to diagnose failed upgrades at the moment, and something which would help us (and the OP of #239) is time-stamping in the log, along with a record of exceptions that have caused update.exe to fail completely.

We had to rearrange things slightly so that the logger is above the exception handler, but other than that it's a very minor mod.  Having two try/catch blocks is unnecessary but I was trying to make the patch as small as possible.  

Date/time stamps are in local time, in a format roughly like ISO8601, to avoid any international date order confusion.


paulcbetts:
Thanks @willdean!

PR #265, opened 2015-03-04
Removed a few unused or unneeded DLL refs from Squirrel.dll

willdean:
Don't care if you don't want this one, but not linking with Windows.Forms.Dll for the sake of one unused enum feels like it might be a good thing


paulcbetts:
Thanks!

PR #266, opened 2015-03-04
Filled in one of the missing integration tests for comment

willdean:
This was my attempt at filling in one of the tests with `Assert.False(true, "Rewrite this to be an integration test");`

Let me know if this is correct/useful.


BarryThePenguin:
Method should use `public async Task` not `public async void` otherwise XUnit won't know when the test has completed

willdean:
@BarryThePenguin Thanks, fixed

PR #276, opened 2015-03-08
Remove unused applicationName field/ctor arg

willdean:
ApplyReleasesImpl has an unused field which does nothing, but does share a name with some method parameters.  This removes it


paulcbetts:
Thanks! :+1:

Issue #280, opened 2015-03-10
CreateShortcutsForExecutable ShortcutLocation.StartMenu puts it in a sub-group


stefanolson:
Currently CreateShortcutsForExecutable when creating a program group puts it inside of a group folder with the name of the author.  This may be what some people want, but I would like my single icon to appear directly in the start menu. Could we have an option for this please?

Thanks,
Stefan

willdean:
@paulcbetts  would the best way to do this be to add another flag in ShortcutLocation (e.g. StartMenuRoot)?
If so, I can do this and put in a PR if you like

stefanolson:
@willdean thanks!

utd1878:
Was there ever a way to ensure the icon was created outside this group folder?

Thieum:
@shiftkey this can be discussed in #484 , like proposed in #281 , so this can be closed.

PR #281, opened 2015-03-10
Added additional shortcut location options

willdean:
Possible fix for #280

### DIFF
--- a/src/Squirrel/IUpdateManager.cs
+++ b/src/Squirrel/IUpdateManager.cs
@@ -13,13 +13,30 @@ namespace Squirrel
 {
     [Flags]
     public enum ShortcutLocation {
+        /// <summary>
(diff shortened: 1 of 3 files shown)


jamesopti:
@willdean Is this still an issue? I'm looking for this workaround and am not sure if this is the only way!

willdean:
@jamesopti I don't know - I just stuck in a PR against an issue which someone else had opened - neither the PR nor the issue have ever gone anywhere, and I don't know if the project has provided this functionality some other way.

Thieum:
@shiftkey maybe this should be revived? It seems to adress a couple of issues.

Thieum:
would fix #424 as well.

shiftkey:
Let's talk about this more in #424, because if the escape hatch for making a custom shortcut isn't satisfactory we should figure out how to address that specific problem.

Issue #284, opened 2015-03-16
You must dispose UpdateManager!


peters:
As of version 0.9.2 squirrel is still leaking a mutex when checking for updates. We have all errors reported to SmartAssembly and the following exceptions are occuring frequently:

![image](https://cloud.githubusercontent.com/assets/869/6662513/9b0e0e2a-cbbc-11e4-8282-c28f3d4fdf74.png)

![image](https://cloud.githubusercontent.com/assets/869/6662515/a308d7e0-cbbc-11e4-8d92-fc67437eeea6.png)

The update code is really simple, i.e:

```cs
using (var updateManager = new UpdateManager(updateUrl, "application", FrameworkVersion.Net45)) {
   
}

updateSuccess(releaseEntry);
```

Might be related to https://github.com/Squirrel/Squirrel.Windows/issues/138 ?

We check for updates every minute.

/cc @paulcbetts

GeertvanHorrik:
Looks like the update.exe (program.cs) already puts all instantiations in usings? Only a few tests (CheckForUpdateTests) are not yet converted.

ayezutov:
Hello @peters, @paulcbetts, Everyone,

I faced the same issue few days back and tried to find a solution to it.

In my particular case replacing
```c#
using (var mgr = new UpdateManager(updateUrl, "X", FrameworkVersion.Net45))
{
      await mgr.CheckForUpdate();
}
```
with
```c#
using (var mgr = new UpdateManager(updateUrl, "X", FrameworkVersion.Net45))
{
      var checkForUpdateTask = mgr.CheckForUpdate();
      checkForUpdateTask.Wait();
}
```
did solve the issue.

Here are couple of debugger screenshots.
In this one the "Next statement" line was never hit:
![image](https://cloud.githubusercontent.com/assets/175635/6982330/5e01904e-d9de-11e4-8eb3-625f9fbeced3.png)



In this one everything is as expected:

![image](https://cloud.githubusercontent.com/assets/175635/6982318/349aaab0-d9de-11e4-8026-e534960ac5dc.png)

mshumilov:
@ayezutov Actually you have have this error in case you will close your app before check for update finished. The solution for this is to make UpdateManager as class property and dispose it on app close.

ayezutov:
@mshumilov Thanks a lot for the answer! It made me revisit the calls once again and I was able to find the missing `await`, so the application indeed exited before dispose was called.

[...]

PR #289, opened 2015-03-19
Dispose the update lock after we've used it

flagbug:
I'm not sure if the update lock should even be disposed, but it seems like it, since the lock file doesn't get deleted otherwise (https://github.com/Squirrel/Squirrel.Windows/blob/dab4d8efafb0144952dc41bcfb800ddeb21cc716/src/Squirrel/Utility.cs#L609)

Fixes https://github.com/Squirrel/Squirrel.Windows/issues/284 maybe?


flagbug:
Wait, nevermind, I'm dumb, the lock goes for the whole `UpdateManager` instance

PR #294, opened 2015-03-25
Include Squirrel.exe in libs

ricardopolo:
This allows to have Squirrel itself always updated, including `Update.exe` like `Squirrel.dll` is already included.

It will be add some size to the first package, but with Delta updates it will not be a problem. IMO the trade-off between a little more size in 'Setup.exe' for having Squirrel always updated worth it.


paulcbetts:
This will cause Squirrel.exe to be added as a reference, which definitely isn't what we want. I'm :+1: on the concept, but we need to patch this into Releasify

paulcbetts:
In the future, just reset the branch, git push -f to it, then keep going! No need to open a new PR :)

PR #295, opened 2015-03-25
Allow to specify Setup exe name

ricardopolo:
If you prefer another name instead "setup.exe" for your installer
capture

And you get this RELEASES folder

capture

This PR also updates all the NuGets packages used by the project.


paulcbetts:
I think we don't need another parameter for this, you can just rename the file

PR #296, opened 2015-03-25
Args of setup

ricardopolo:
This PR is because I need some help

Using this logic I (think) can pass arguments from setup.exe to invokePostInstall method. But I dont know how can I read it from install event.

Any help will be appreciated.


ricardopolo:
@paulcbetts I think this ir ready and you can read arguments in your exe

Issue #288, opened 2015-03-19
UpdateManager.DownloadReleases does not report progress in %


ogborstad:
According to the API documentation the progress parameter in the callback is supposed to be between 0-100. I am getting large numbers which appear to be downloaded bytes...

Cyral:
The same happens with `UpdateApp`, it returns some very large numbers while downloading before reporting the progress of installation.

paulcbetts:
This is definitely a bug, I think this was introduced with one of @sickboy's PRs - cc /@sickboy

sickboy:
Whoops! How does that look: #297

Thieum:
@shiftkey fixed in #297 , this can be closed.

PR #297, opened 2015-03-26
Fix progress component

sickboy:
Fix for #288


ogborstad:
Better, but for 3 downloads it will never report progress 100% due to 
each component will max out at 33 and 33 + 33 + 33 = 99

sickboy:
@ogborstad fair enough, adjusted!

PR #308, opened 2015-04-14
Install to ProgramData when LocalAppData is broken

paulcbetts:
For a dizzying array of reasons, installing anything to `%LocalAppData%` simply doesn't work, or throws bizarre errors. It turns out though, that since nobody knows what the hell to do with `C:\ProgramData`, nobody looks there either. When installations fail to `%LocalAppData%`, we're gonna fall back to `C:\ProgramData\$PackageName`. On Windows 7 where `C:\ProgramData` doesn't exist, we'll just make it.

This PR also adds an `IsInstalledApp` property on UpdateManager which lets you tell if the application is currently installed.

## Pay Attention To This!

If your application currently is hard-coded to look in `%LocalAppData%` for stuff, this PR will break you (well technically, these users were never able to install in the first place, but now they will be able to and now you're gonna see weirdness). Instead, use UpdateManager's `RootAppDirectory` property, or do all of your operations relative to `Assembly.GetExecutingAssembly().Location`.

## TODO

- [x] Remove our use of LocalAppData in Update.exe
- [x] Rig Setup to fall back to ProgramData
- [x] Remove LocalAppData from Squirrel
- [x] Make sure this works everywhere

### DIFF
--- a/src/Setup/UpdateRunner.cpp
+++ b/src/Setup/UpdateRunner.cpp
@@ -140,7 +140,7 @@ int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallback
 	if (!useFallbackDir) {
 		SHGetFolderPath(NULL, CSIDL_LOCAL_APPDATA, NULL, SHGFP_TYPE_CURRENT, targetDir);
 	} else {
-		ExpandEnvironmentStrings(targetDir, L"%HOMEDRIVE%\\ProgramData", _countof(targetDir));
+		ExpandEnvironmentStrings(L"%HOMEDRIVE%\\ProgramData", targetDir, _countof(targetDir));
 
 		// NB: HOMEDRIVE may be unset or jacked up
 		if (targetDir[0] == L':') {
--- a/src/Setup/UpdateRunner.cpp
+++ b/src/Setup/UpdateRunner.cpp
@@ -141,6 +141,12 @@ int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallback
 		SHGetFolderPath(NULL, CSIDL_LOCAL_APPDATA, NULL, SHGFP_TYPE_CURRENT, targetDir);
 	} else {
 		ExpandEnvironmentStrings(targetDir, L"%HOMEDRIVE%\\ProgramData", _countof(targetDir));
+
+		// NB: HOMEDRIVE may be unset or jacked up
+		if (targetDir[0] == L':') {
+			wcscpy_s(targetDir, _countof(targetDir), L"C:\\ProgramData");
+		}
+
 		if (!CreateDirectory(targetDir, NULL) && GetLastError() != ERROR_ALREADY_EXISTS) {
 			wchar_t err[4096];
 			_swprintf_c(err, _countof(err), L"Unable to write to %s - IT policies may be restricting access to this folder", targetDir);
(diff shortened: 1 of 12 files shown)


src/Setup/UpdateRunner.cpp
```diff
@@ -137,13 +137,34 @@ int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine)
 	wchar_t logFile[MAX_PATH];
 	std::vector<CString> to_delete;
 
-	SHGetFolderPath(NULL, CSIDL_LOCAL_APPDATA, NULL, SHGFP_TYPE_CURRENT, targetDir);
+	if (!useFallbackDir) {
+		SHGetFolderPath(NULL, CSIDL_LOCAL_APPDATA, NULL, SHGFP_TYPE_CURRENT, targetDir);
+	} else {
+		ExpandEnvironmentStrings(L"%HOMEDRIVE%\\ProgramData", targetDir, _countof(targetDir));
```
CharlieHess: Isn't this just `CSIDL_COMMON_APPDATA`? (Older versions use C:\Documents and Settings\All Users\Application Data).

src/Squirrel/UpdateManager.cs
```diff
@@ -190,6 +194,33 @@ public static void RestartApp(string exeToStart = null, string arguments = null)
...
+                // dunno lol
+                return Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+            }
+
+            assemblyLocation = assemblyLocation ?? assembly.Location;
+
+            if (Path.GetFileName(assemblyLocation).Equals("update.exe", StringComparison.OrdinalIgnoreCase)) {
+                // NB: Both the "SquirrelTemp" case and the "App's folder" case 
+                // mean that the root app dir is one up
+                var oneFolderUpFromAppFolder = Path.Combine(Path.GetDirectoryName(assemblyLocation), "..");
```
CharlieHess: How about `Directory.GetParent` instead?

paulcbetts: Legit

src/Setup/UpdateRunner.cpp
paulcbetts: Probably a better idea

sandyarmstrong:
Are there any issues filed describing any of these "dizzying array of reasons"? I'd like to understand this better before committing to a local user install approach.

paulcbetts:
>  I'd like to understand this better before committing to a local user install approach.

Strangely enough, none of these issues are filed on Squirrel - I suspect that Squirrel hasn't gotten enough Traction In The Enterprise(tm) yet, but we definitely see it on Atom and in the Slack Windows app. Apparently a lot of IT shops have followed the steps in http://www.computerworld.com/article/2485214/microsoft-windows/cryptolocker-how-to-avoid-getting-infected-and-what-to-do-if-you-are.html which ends up blocking anything in AppData. Others have completely borked permissions on AppData because Security(tm).

sandyarmstrong:
...wow. This sounds like it's going to need to evolve into a "your company fucked up permissions, where would you like to install?" dialog, ultimately.

paulcbetts:
> ...wow. This sounds like it's going to need to evolve into a "your company fucked up permissions, where would you like to install?" dialog, ultimately.

Well, so this PR _does_ mostly set us up to run anywhere, including _(vomits in mouth)_ Program Files. After this PR, all we need to do is detect when we need to elevate and rerun ourselves.

stajs:
> Strangely enough, none of these issues are filed on Squirrel - I suspect that Squirrel hasn't gotten enough Traction In The Enterprise(tm) yet

Sadly, I work at an Enterprise(tm) place and yep, our Group Policy blocks executables in `%LocalAppData%`. I didn't file an issue about it, because TBH it is the fault of the Enterprise not Squirrel. I did however ask that `/docs/troubleshooting.md` be updated with a note about this for other Enterprise users.

BarryThePenguin:
image

PR #311, opened 2015-04-23
Fail with a nice error message if running on XP

JohnThomson:
(or on various other OS's that can't install .NET 4.5)
Using PlatformToolset v120_xp allows Setup.exe itself to run on XP,
since it is already a 32-bit application. But installing the service pack
will fail quietly, so give a nice message rather than attempting it.

### DIFF
--- a/src/Setup/FxHelper.cpp
+++ b/src/Setup/FxHelper.cpp
@@ -6,6 +6,19 @@
 static const wchar_t* ndpPath = L"SOFTWARE\\Microsoft\\NET Framework Setup\\NDP\\v4\\Full";
 static const int fx45ReleaseVersion = 378389;
 
+// According to https://msdn.microsoft.com/en-us/library/8z6watww%28v=vs.110%29.aspx,
+// to install .NET 4.5 we must be Vista SP2+, Windows 7 SP1+, or later.
+bool CFxHelper::CanInstallDotNet4_5()
+{
+	if (!IsWindowsVistaSP2OrGreater())
+		return false;
+	if (IsWindows7SP1OrGreater())
+		return true;
+	// At this point, we might be: an acceptable version of Vista, or an unacceptable version of Win7.
+	// Being windows 7 is the remaining failure case.
+	return !IsWindows7OrGreater();
+}
+
 bool CFxHelper::IsDotNet45OrHigherInstalled()
 {
 	ATL::CRegKey key;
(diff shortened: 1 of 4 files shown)


src/Setup/winmain.cpp
```diff
@@ -23,6 +23,14 @@ int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
 	CString cmdLine(lpCmdLine);
 	bool isQuiet = (cmdLine.Find(L"-s") >= 0);
 
+	if (!CFxHelper::CanInstallDotNet4_5())
+	{
```
paulcbetts: Throw this bracket on the line above it

src/Setup/winmain.cpp
```diff
@@ -23,6 +23,14 @@ int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
...
+		// Explain this as nicely as possible and give up.
+		MessageBox(0L, L"This program cannot run on this sytem; it requires Windows Vista SP2 or Windows 7 SP1 or a later version.", L"Incompatible Operating System", 0);
```
paulcbetts: Just say "Windows 7 or Higher", if they're on Vista they won't see this message anyways

src/Setup/FxHelper.cpp
```diff
@@ -6,6 +6,19 @@
 static const wchar_t* ndpPath = L"SOFTWARE\\Microsoft\\NET Framework Setup\\NDP\\v4\\Full";
 static const int fx45ReleaseVersion = 378389;
 
+// According to https://msdn.microsoft.com/en-us/library/8z6watww%28v=vs.110%29.aspx,
+// to install .NET 4.5 we must be Vista SP2+, Windows 7 SP1+, or later.
+bool CFxHelper::CanInstallDotNet4_5()
+{
+	if (!IsWindowsVistaSP2OrGreater())
+		return false;
+	if (IsWindows7SP1OrGreater())
```
paulcbetts: Even though it's not exactly perfect, can you just simplify the condition to >= Win7?

paulcbetts:
This looks great! Think we can just tweak a few things and :shipit:

src/Setup/FxHelper.cpp
JohnThomson: Done

src/Setup/winmain.cpp
JohnThomson: Sorry, should have seen that was your coding standard. Done.

src/Setup/winmain.cpp
JohnThomson: Done.

JohnThomson:
I think I've done the things you asked for...anything else you want changed?

paulcbetts:
Oops, sorry I missed this, I'll have a look again soon

PR #312, opened 2015-04-24
Update.exe: take OrderByDescending to find the latest version of the app directory

PapaMufflon:
`update.exe --processStart MyApplicationName.exe` should search for the latest app directory and then start the exe in there. Currently, the method uses `OrderBy` to order the directories by version and `FirstOrDefault` to get the latest app directory.
I think this is wrong, as this will get you the oldest version and therefore the oldest app directory. Use `OrderByDescending` to fix this.


paulcbetts:
Thanks @papamufflon!

PR #325, opened 2015-05-04
Starting process with proper working dir

### LINKED ISSUE #321: Could not load file or assembly NuGet.Core
Hello.

We are recently having this exeption while applying updates:

`Could not load file or assembly 'NuGet.Core, Version=2.8.50926.602, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)`

I checked and in the App folder exists this DLL
image

And in the App config we have this binding redirection

````xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <runtime>
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="NuGet.Core" publicKeyToken="31bf3856ad364e35" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-2.8.60318.667" newVersion="2.8.60318.667" />
      </dependentAssembly>
````

Any idea about what more should we check?
I think this is related with the latest Nuget update, without modifying the code it was working some days ago.

Thank you

distantcam:
Hopefully fixes #321

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -450,7 +450,7 @@ namespace Squirrel.Update
 
             try {
                 this.Log().Info("About to launch: '{0}': {1}", targetExe.FullName, arguments ?? "");
-                Process.Start(new ProcessStartInfo(targetExe.FullName, arguments ?? ""));
+                Process.Start(new ProcessStartInfo(targetExe.FullName, arguments ?? "") { WorkingDirectory = Path.GetDirectoryName(targetExe.FullName) });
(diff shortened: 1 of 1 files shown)


paulcbetts:
Into it, though the fact that these apps / the framework even cares is pretty suck :(

PR #330, opened 2015-05-06
NuGet.core updated to 2.8.5

### LINKED ISSUE #321: Could not load file or assembly NuGet.Core

ricardopolo:
While fixing  #321 I see that Squirrel (update.exe) was not the reading the binding redirection in my App.config for Nuget.Core 2.8.5, but still looking for the 2.8.3 version.

The Squirrel Nuget installs Nuget.Core 2.8.5 but the Squirrel source code uses Nuget.Core 2.8.3. Thas why I do belive that is still reading that old version that is not in the Bin folder.

This PR simply update Nuget.Core to 2.8.5 version in the source code to match the version used in Squirrel Nuget.

For fix #321 I had to downgrade the Nuget.Core version and now it works, the PR purposed did not fix it.

### DIFF
--- a/src/Squirrel/Squirrel.csproj
+++ b/src/Squirrel/Squirrel.csproj
@@ -63,9 +63,9 @@
     <Reference Include="Mono.Cecil.Rocks">
       <HintPath>..\..\packages\Mono.Cecil.0.9.5.4\lib\net40\Mono.Cecil.Rocks.dll</HintPath>
     </Reference>
-    <Reference Include="NuGet.Core, Version=2.8.50926.602, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+    <Reference Include="NuGet.Core, Version=2.8.60318.667, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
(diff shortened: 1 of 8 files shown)


paulcbetts:
Thanks!

ricardopolo:
@paulcbetts would you bump a new Nuget release please? :shipit:

PR #331, opened 2015-05-06
Update dependencies and handle octokit upgrade

### LINKED ISSUE #329: Update not running after install, works from visual studio
For the past couple months, I've been having no issues with Squirrel.Windows, but recently (can't say for sure when it started, but I'm on 0.9.4 now), the application no longer updates after being installed.

When running from Visual Studio (in Release or Debug) it works fine.  The code in use is: https://github.com/amweiss/vigilant-cupcake/blob/master/VigilantCupcake/MainForm.cs#L507 which is nearly a copy/paste from the example posted here.  Any ideas what could cause the difference?  I'm trying to find a good way to attach a debugger and get symbols working post-install.

Fiddler shows the request going out to the releases URL, and no exceptions are bubbling up.

amweiss:
Hopefully fixes #329

### DIFF
--- a/.nuget/packages.config
+++ b/.nuget/packages.config
@@ -1,7 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="ILRepack" version="1.25.0" />
   <package id="ILRepack" version="1.26.0" />
-  <package id="xunit.runners" version="2.0.0-rc3-build2880" />
   <package id="xunit.runners" version="2.0.0" />
 </packages>
--- a/.nuget/packages.config
+++ b/.nuget/packages.config
@@ -1,5 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="ILRepack" version="1.25.0" />
+  <package id="ILRepack" version="1.26.0" />
   <package id="xunit.runners" version="2.0.0-rc3-build2880" />
+  <package id="xunit.runners" version="2.0.0" />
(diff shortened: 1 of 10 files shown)


src/SyncReleases/SyncImplementations.cs
```diff
@@ -50,7 +50,7 @@ internal class SyncImplementations
 
             await releases.ForEachAsync(async release => {
                 // NB: Why do I have to double-fetch the release assets? It's already in GetAll
-                var assets = await client.Release.GetAssets(nwo.Item1, nwo.Item2, release.Id);
+                var assets = await client.Release.GetAllAssets(nwo.Item1, nwo.Item2, release.Id);
```
amweiss: This is due to https://github.com/octokit/octokit.net/releases/tag/v0.10.0

.nuget/packages.config
```diff
@@ -1,5 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="ILRepack" version="1.25.0" />
+  <package id="ILRepack" version="1.26.0" />
```
flagbug: Seems like there's gone something wrong here

amweiss: Hm, good call, let me go try to fix that

amweiss:
As a note, after building the library with these changes, I dropped the .dlls into an installed app and it started working again, so hopefully this covers the current issues with strong naming.

paulcbetts:
Thanks @amweiss

PR #337, opened 2015-05-09
Per-Machine installs

### LINKED ISSUE #314: Installer doesn't offer an option to install for all users
In an enterprise environment, it is desirable to have apps installed in Program Files as part of a standard computer or app deployment process. Google uses %LocalAppData% for Chrome but offers an option for enterprise installation.
Could we get a silent install switch that installs silently in Program Files instead?

Thanks

paulcbetts:
This PR adds a new `--machine` parameter that will install the app for all users (they may need to log out and log back in)

## TODO:

- [x] Verify this works correctly to install on the current profile
- [x] Verify this installs on a current user that isn't the current profile
- [x] Verify this installs on a newly created user
- [x] Uninstall detritus gets detected as valid install
- [x] Silent doesn't get passed to Update.exe
- [x] Installing as admin should install for admin's profile

Fixes #314

### DIFF
--- a/src/Setup/MachineInstaller.cpp
+++ b/src/Setup/MachineInstaller.cpp
@@ -26,7 +26,6 @@ bool findPackageFromEmbeddedZip(wchar_t* buf, DWORD cbSize)
 	int index = 0;
 	do {
 		ZIPENTRY zentry;
-		wchar_t targetFile[MAX_PATH];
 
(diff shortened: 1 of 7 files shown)


Neustradamus:
All good softwares when we install use by default "Program Files" (or recently "Program Files (x86)" for 32 bit apps in 64 bit system).
Bad softwares do not use it.

It is since Windows 95 and maybe before, no?

PR #343, opened 2015-05-14
Improve update progress

JohnThomson:


JohnThomson:
Apologies...this was not meant to target the original fork at all.

PR #344, opened 2015-05-14
Make download progress reach 100% reliably

JohnThomson:
FileDownloader.DownloadFile ignores progress notifications received less than 500ms after the previous notification. This means that a random amount of progress, up to half a second per file, is lost from the cumulative total. We've seen updates report as little as 40% complete by the time they were entirely done. This commit updates the progress correctly at the end of each download.

### DIFF
--- a/src/Squirrel/UpdateManager.DownloadReleases.cs
+++ b/src/Squirrel/UpdateManager.DownloadReleases.cs
@@ -40,6 +40,16 @@ namespace Squirrel
                                 progress((int)Math.Round(current += component));
                             }
                         });
+                        // With a lot of small updates, and since notifications are only sent every half second
+                        // for each one, we can easily miss half or more of the progress on each download.
+                        // To make sure we eventually get to 100% of the whole process, we need to update
+                        // progress (and especially the total in current) to indicate that this one is complete.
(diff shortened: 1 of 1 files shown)


src/Squirrel/UpdateManager.DownloadReleases.cs
```diff
@@ -40,6 +40,16 @@ public DownloadReleasesImpl(string rootAppDirectory)
                                 progress((int)Math.Round(current += component));
                             }
                         });
+                        // With a lot of small updates, and since notifications are only sent every half second
+                        // for each one, we can easily miss half or more of the progress on each download.
+                        // To make sure we eventually get to 100% of the whole process, we need to update
+                        // progress (and especially the total in current) to indicate that this one is complete.
+                        lock (progress)
+                        {
```
paulcbetts: :arrow_up:

paulcbetts:
tbh I'm not _super_ excited to touch this code, people keep messing with it and every time we end up shipping a regression

JohnThomson:
Up to you, we have it in our fork and it seems to be an improvement. Just trying to do our Open Source duty and contribute back to the original.

PR #345, opened 2015-05-15
Use class member applicationName variable instead of local in UpdateManager's constructor.

theofanis:
I keep getting an exception from Path.Combine after updating to 0.99 if applicationName parameter is null, as suggested since it's now optional.

I guess this is just a small oversight and the local applicationName is used to find the rootAppDirectory, even though the class member applicationName is correctly set beforehand.


paulcbetts:
hwhoops, thanks!

PR #348, opened 2015-05-18
Retry downloads

JohnThomson:
This is especially helpful in low-bandwidth situations, and even more so if bandwidth is expensive.

### DIFF
--- a/src/Squirrel/UpdateManager.DownloadReleases.cs
+++ b/src/Squirrel/UpdateManager.DownloadReleases.cs
@@ -40,6 +40,16 @@ namespace Squirrel
                                 progress((int)Math.Round(current += component));
                             }
                         });
+                        // With a lot of small updates, and since notifications are only sent every half second
+                        // for each one, we can easily miss half or more of the progress on each download.
+                        // To make sure we eventually get to 100% of the whole process, we need to update
+                        // progress (and especially the total in current) to indicate that this one is complete.
+                        lock (progress)
+                        {
+                            current -= component;
+                            component = toIncrement;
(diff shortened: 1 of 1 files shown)


src/Squirrel/UpdateManager.DownloadReleases.cs
```diff
@@ -19,29 +20,59 @@ public DownloadReleasesImpl(string rootAppDirectory)
...
-                            lock (progress) {
-                                current -= component;
-                                component = toIncrement / 100.0 * p;
-                                progress((int)Math.Round(current += component));
-                            }
-                        });
-                    });
-                } else {
+                    Exception lastException = null;
+                    for (int attempts = 0; ; attempts++) {
```
paulcbetts: Check the Utility class, there's already a Retry method

JohnThomson: Interesting idea, and I had missed that method, but retrying for every possible exception may not be optimal here. More significantly, completing the block does not amount to success...success here is after the initial test that all required files have been correctly downloaded. Nor is it good to move that test to the end...then we lose not downloading anything we already have. Then I ran into the complication that C# won't let me use await in a block that is passed to Retry...I think trying to use it adds more complication than it removes.

anonymate:
Is this code for updating / download of the software, right?, type once did I do an autopatch for a MMORPG and I still have the code here, am going to give one analysed in him ^^, any more I if am not wrong he shows to be more dynamic than that one, so that I will test it I will be going to bring it here the one who knows the code does not finish helping in anything is not true?

PR #353, opened 2015-05-21
ILMerge all Strongly Named binaries

### LINKED ISSUE #350: App not updating in squirrel.dll v0.99.0.0
Hi

I was previously using squirrel.dll v0.9.4.0 and it was working fine, but now after the nuget package update to .99 app is not updating, inside the appData\Local\AppName\packages the nupkg and delta files are created for the update but no entries added for in RELEASE file

i am calling update as below

 Task.Run(async () =>
            {
                using (var mgr = new UpdateManager(@"Path/to/Update/Url", "AppName"))
                {
                    await mgr.UpdateApp();
                }
            });

am I missing something?

### LINKED ISSUE #351: Nuget.core version is 2.8.60318.667 in 0.99
As indicated in issue #350, the nugget.core version is now 2.8.60318.667. What appears to have happened is that the dependencies have been updated but the dependency requirements in the nuget package have not been updated (this also applies to splat and mono.cecil).

The problem that I discovered was that I was getting error reports like this:
Exception]:[System.IO.FileLoadException: Could not load file or assembly 'NuGet.Core, Version=2.8.60318.667, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040) File name: 'NuGet.Core, Version=2.8.60318.667, Culture=neutral, PublicKeyToken=31bf3856ad364e35'

Because the nuget package for nugget.core had not been updated to the same as the version used by squrriel.  Once I updated that and splat and mono.cecil updates work. The problem that I have is that there are some versions out in the wild where updates aren't working because they have the old nugget.core that is not the same as the version used by squrriel.  Is there any way around this?

Thanks,
Stefan

paulcbetts:
This PR fixes The NuGet Problem by just ILMerging it so we don't have to think about Strong Naming and its dumb problems:

Fixes #350 
Fixes #351
Fixes #352

### DIFF
--- a/src/Squirrel.nuspec
+++ b/src/Squirrel.nuspec
@@ -9,9 +9,7 @@
 
     <dependencies>
       <dependency id="DeltaCompressionDotNet" version="1.0.0" />
-      <dependency id="NuGet.Core" version="2.8.3" />
-      <dependency id="Mono.Cecil" version="0.9.5.4" />
-      <dependency id="Splat" version="1.5.1" />
+      <dependency id="Splat" version="1.6.2" />
     </dependencies>
 
     <id>squirrel.windows</id>
--- a/src/Squirrel/Squirrel.csproj
+++ b/src/Squirrel/Squirrel.csproj
@@ -55,18 +55,6 @@
       <HintPath>..\..\packages\Mono.Cecil.0.9.6.1\lib\net45\Mono.Cecil.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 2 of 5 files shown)


ricardopolo:
Does it mean that we dont need dependencies like Nuget.Core anymore?

:)

paulcbetts:
@ricardopolo Yep! Just remove it

PR #356, opened 2015-05-21
Repair ILRepack command to produce correct assembly name

JohnThomson:
/out: command must have desired assembly name as file name before extension

(The old version produced a squirrel with squirrel.dll as the assembly name, and then when code tries to load squirrel.dll, the loader looks for squirrell.dll.dll and fails to find it, and the client program won't run.)


paulcbetts:
:+1:, thanks for the great work to track this down!

PR #363, opened 2015-06-05
Display "Open Setup Log" button when there IS an install failure log

JohnThomson:
The old version shows the "Open Setup Log" button
exactly when it should be hidden (because there is none)


paulcbetts:
D'oh, thanks!

PR #387, opened 2015-07-16
Update ACL to work with Windows 7.  AC SID only valid on Windows 8+

allanrsmith:
When comparing the SquirrelMachineInstalls SDDL to Program Files I do notice that the AI (Auto Inherit) DACL flag is removed.  This could be due to the line: ttrs.bInheritHandle = false;

Leaving as is for now since I am not sure what the intention was.


paulcbetts:
Thanks for the great writeup and PR! @allanrsmith

PR #394, opened 2015-07-23
Decode/encode entry names of nupkg files to prevent PathTooLongException

duanyao:
If some files/dirs to be installed contains non-ASCII chars, nuget will %-encode them in `nupkg` file, which may result in very long path and causes `PathTooLongException` while extracting to disk.

For example, one Chinese char is encoded as 3 bytes in UTF-8, and each byte is encoded as 3 chars in %-encoding, so one Chinese char is expanded to 9 chars in %-encoding, which causes `PathTooLongException` too easily (the limit is only 260 chars).

This patch decode/encode %-encoded entry names during creation of installation package to prevent this issuse.


paulcbetts:
Sorry for the huge delay, this is great!

Issue #389, opened 2015-07-19
Signing of Setup.exe fails through Squirrel


Pinpickle:
When using `--releasify` with `-n` and a Digicert certificate, I get an error. It signs the executables inside the package just fine, but the final installer file, `Setup.exe`, has an error thrown when Squirrel tries to sign it

```
squirrel : System.AggregateException: One or more errors occurred. ---> System.Exception: Failed to sign, command invoked was: [...]\packages\squirrel.windows.0.99.1.1\tools\signtool.exe sign /t 
http://timestamp.digicert.com [...]\Releases\Setup.exe'
```

Full stack trace:

```
   at Squirrel.Update.Program.<signPEFile>d__93.MoveNext()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at Squirrel.Update.Program.Releasify(String package, String targetDir, String packagesDir, String bootstrapperExe, String backgroundGif, String signingOpts, String baseUrl, String setupIcon)
   at Squirrel.Update.Program.executeCommandLine(String[] args)
   at Squirrel.Update.Program.main(String[] args)
   at Squirrel.Update.Program.Main(String[] args)
---> (Inner Exception #0) System.Exception: Failed to sign, command invoked was: '[...]\pcr-touch\packages\squirrel.windows.0.99.1.1\tools\signtool.exe sign /t http://timestamp.digicert.com 
[...]\Projects\pcr-touch\Releases\Setup.exe'
   at Squirrel.Update.Program.<signPEFile>d__93.MoveNext()<---
```

I can get around it at the moment by signing the file manually, and the command that is specified that fails with Squirrel runs successfully when I run it manually (with quotes around paths).

There's no more information other than the stack trace in the Squirrel log file.

naishx:
I had the same issue this is because you have blanks in the outpath?

The root of the problem:

Tuple<int, string> processResult = await Utility.InvokeProcessAsync(exe,
                String.Format("sign {0} {1}", signingOpts, exePath), CancellationToken.None);

should be changed to this:

Tuple<int, string> processResult = await Utility.InvokeProcessAsync(exe,
[...]

PR #402, opened 2015-08-04
Fixed signtool error if outpath contains blanks.

naishx:
Issue:  Signing of Setup.exe fails through Squirrel #389


paulcbetts:
Perfect! Thanks for contributing!

Issue #404, opened 2015-08-07
No JSON output from --download


DDRBoxman:
Seems like Console.WriteLine just isn't working. I can pass --update and the update works successfully though.

https://github.com/Squirrel/Squirrel.Windows/blob/master/src/Update/Program.cs#L149

```
2015-08-07 15:58:41> Program: Starting Squirrel Updater: --download=http://localhost:8000
2015-08-07 15:58:41> Program: Fetching update information, downloading from http://localhost:8000
2015-08-07 15:58:41> CheckForUpdateImpl: Downloading RELEASES file from http://localhost:8000
2015-08-07 15:58:41> FileDownloader: Downloading url: http://localhost:8000/RELEASES
2015-08-07 15:58:42> FileDownloader: Downloading file: http://localhost:8000/Muxy-Ticker-0.1.4-full.nupkg
2015-08-07 15:58:42> UpdateInfo: Couldn't get release notes for:Muxy-Ticker-0.1.4-full.nupkg: System.Exception: Invalid 'ReleaseNotes' value in nuspec file at 'C:\Users\UncodinDev\AppData\Local\Muxy\packages\Muxy-Ticker-0.1.4-full.nupkg'
   at Squirrel.ReleaseEntry.GetReleaseNotes(String packageDirectory)
   at Squirrel.UpdateInfo.<FetchReleaseNotes>b__2(ReleaseEntry x)
```

DDRBoxman:
Current observed behavior is that the command seems to instantly finish with no time to return anything.
But the entire process keeps going in the background and does generate the expected results if I show it in a MessageBox.

No luck tracking down why the process is disconnecting from the command line and still running yet.

DDRBoxman:
I'm adding a gif here to hopefully better illustrate what I'm seeing.

![9132e88193586b56d0d6f2a2af2940eb](https://cloud.githubusercontent.com/assets/207897/9183679/45f81b00-3f76-11e5-9f35-0ab0dc8c724b.gif)

The app does run correctly as shown by the MessageBox but nothing ever makes it to stdout.

DDRBoxman:
Apparently the answer is to use Update.com that was added here: https://github.com/Squirrel/Squirrel.Windows/issues/27

Seems like this might help with some of the windows update issues people are having with Atom too: https://github.com/atom/atom/blob/master/src/browser/squirrel-update.coffee#L8

paulcbetts:
If you run Update directly, you need the com shim

[...]

PR #408, opened 2015-08-11
Package Update.com in the installer and install it

DDRBoxman:


DDRBoxman:
#404 shows how I ended up here.

paulcbetts:
This is only necessary when you run Update.exe directly from console - since callers will always be capturing output via pipes, this isn't necessary

PR #412, opened 2015-08-17
Reworked steps order, and image inclusion

Thieum:
The image describing the package was not at the right place.
The step providing code to check for update, should be done before packaging the files.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -2,10 +2,6 @@
 
 * `Install-Package squirrel.windows`, or grab the binaries from the Releases section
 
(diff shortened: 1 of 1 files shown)


paulcbetts:
@Thieum Thanks!

Thieum:
Considering the comment about startup, I tried to put the documented code in the Application_Exit event handler, but it never triggered the update. I tried the different syntaxes, async:
            Task.Run(async () =>
            {
            using (var mgr = new UpdateManager(Settings.Default.UpdatePath))
            {
                await mgr.UpdateApp();
            }
            }

and non-sync:
            using (var mgr = new UpdateManager(Settings.Default.UpdatePath))
            {
                mgr.UpdateApp();
            }
Is it a known limitation?

PR #413, opened 2015-08-18
Improved code example

Thieum:
Improved code example based on my experience, implementing it in a wpf app.


paulcbetts:
Hm, I'm not sure the Task.Run is a good idea here, I think developers should decide that. Thanks though!

PR #430, opened 2015-09-06
Set default/optional LegalCopyright for Setup.exe from nuspec/nupkg

mnemotiv:
This seems obsolete:
- "LegalCopyright", "Copyright (C) 2014"... https://github.com/Squirrel/Squirrel.Windows/blob/master/src/Setup/Setup.rc#L99

Maybe lets set default new copyright with first owner or get it from nuspec/nupkg.


mnemotiv:
@paulcbetts

paulcbetts:
Into it!

PR #434, opened 2015-09-09
FileDownloader constructor takes optional WebClient as a parameter

ionoy:
Makes it possible to provide custom Credentials and stuff

### DIFF
--- a/src/Squirrel/FileDownloader.cs
+++ b/src/Squirrel/FileDownloader.cs
@@ -13,42 +13,49 @@ namespace Squirrel
 
     class FileDownloader : IFileDownloader, IEnableLogger
     {
+        private readonly WebClient _providedClient;
+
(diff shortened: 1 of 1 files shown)


paulcbetts:
Into it, but what's with all the whitespace changes?

ionoy:
Method body was one tab to the left. I can revert if it was intentional.

paulcbetts:
How in the world did that happen? Your version is definitely Better, merged!

Thieum:
Does it handle the case where the url requires Authentication? That would be perfect for a use case I have :)

grandtiger:
I ran into a use case where URL requires authentication. Is it possible to do? Thanks a lot!

grandtiger:
@ionoy Hi Mihhail, how do you make use of this additional WebClient parameter? Is it possible to leverage this to handle URL requires authentication? Thanks!

PR #447, opened 2015-09-28
Fix undocumented Notification Area registry key on Upgrade

paulcbetts:
This resolves an issue with Squirrel.Windows applications, where if you use `Shell_NotifyIcon` to show a tray icon, every update will put a new entry in the "Always Show" notification preferences. https://github.com/atom/electron/issues/2468#issuecomment-142684129 has the entire sad story on what's happening here.

This PR also exposes a new `TrayStateChanger.PromoteTrayItem` method, which will allow you to promote your tray icon to be always visible unless the user explicitly says "Never show it ever".

### DIFF
--- a/src/Squirrel/Squirrel.csproj
+++ b/src/Squirrel/Squirrel.csproj
@@ -20,6 +20,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -28,6 +29,7 @@
     <DefineConstants>TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'CIBuild|AnyCPU'">
     <OutputPath>bin\CIBuild\</OutputPath>
@@ -89,6 +91,7 @@
     <Compile Include="SquirrelAwareApp.cs" />
     <Compile Include="SquirrelAwareExecutableDetector.cs" />
     <Compile Include="TaskbarHelper.cs" />
+    <Compile Include="TrayHelper.cs" />
     <Compile Include="UpdateInfo.cs" />
     <Compile Include="UpdateManager.CheckForUpdates.cs" />
     <Compile Include="UpdateManager.cs" />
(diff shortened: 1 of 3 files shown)


wearhere:
@paulcbetts is the effect of that `PromoteTrayItem` method to move an item from the customization tray

screen shot 2015-12-30 at 1 52 00 pm

to the toolbar?

screen shot 2015-12-31 at 1 25 21 pm

If so, is there a way to call it from Electron i.e. via one of the Squirrel executables? It would work around https://github.com/atom/electron/issues/3970 (and it'd be nice to show my tray items to the user too).

paulcbetts:
You'd have to use edge.js to do it, or write a small helper exe in C#

wearhere:
Cool, thanks! edge.js looks cool. I assume I can extract `PromoteTrayItem` into its own file and then just copy the import list from the top of `TrayHelper.cs` as DLLs to reference.

paulcbetts:
@wearhere Yeah, that's actually probably the easiest way - use the `edge-atom-shell` package which I wrote specifically for Electron

wearhere:
Awesome thanks!

wearhere:
NPM package [here](https://www.npmjs.com/package/electron-promote-windows-tray-items):

```js
var promoteWindowsTrayItems = require('electron-promote-windows-tray-items');

// Create a tray item.
var icon = new Tray(/* ... */);

if (process.platform === 'win32') {
  // Icon will now be in the toolbar customization area without the user explicitly toggling it to show in the toolbar.

  promoteWindowsTrayItems(function(err) {
    // Icon will now be in the toolbar itself unless the user explicitly hid it from the toolbar.
  });
}
```

Thanks again @paulcbetts!

paulcbetts:
@wearhere Great work!

wearhere:
@paulcbetts do you use an asar archive with edge.js and if so how do you unpack your app as appropriate to avoid the following?

screen shot 2016-01-05 at 8 56 31 pm

I've tried passing `**/*.@(node|dll)` for `--unpack` and no dice.

paulcbetts:
@wearhere That was one of the hacks I had to do to fix edge.js in the `edge-atom-shell` module, yeah

wearhere:
@paulcbetts hmm… ok, so, as you may have guessed/seen by looking at `electron-promote-windows-tray-items`, I didn't use `edge-atom-shell` but rather [`electron-edge`](https://www.npmjs.com/package/electron-edge). I did this because I ran into issues using `edge-atom-shell`—which I can't quite remember, unfortunately, except for that I tried using `electron-rebuild` and it didn't help. It turned out that `electron-edge` didn't require using `electron-rebuild` at all which was really nice.

But now it looks like `electron-edge` doesn't have your asar patch, so… to the best of your knowledge:
* does `edge-atom-shell` work with Electron 0.36.2?
* does `edge-atom-shell` require running `electron-rebuild`?

The error I'm getting now, using `edge-atom-shell` `5.0.0` with `electron-rebuild` `1.0.1` and both Electron `0.36.1` and `0.36.2`, is

screen shot 2016-01-05 at 10 17 06 pm

when trying to open the app. No errors during the build process. Thanks for your help!

paulcbetts:
> does edge-atom-shell work with Electron 0.36.2?

100% sure yes

> does edge-atom-shell require running electron-rebuild?

Yeah.

> %1 is not a valid Win32 application

This means that you're trying to load a 64-bit DLL into a 32-bit app or vice versa, which might hint that electron-rebuild isn't running? Right now electron-rebuild is a garbage fire, I really need to fix it

Issue #449, opened 2015-09-28
Allow prerelease strings in package version numbers


maxbrunsfeld:
We'd like to use versions like `1.1.0-beta.3` in our app. Currently, this doesn't work because Squirrel uses the prerelease field in the version number for `delta` and `full`.

nathansobo:
> and old clients MUST be able to upgrade

For our beta channel, we'll have users download a new installer. For our stable channel, we'll continue to use un-suffixed release names. So it seems like there will be continuity where it's needed unless I'm missing something.

paulcbetts:
> So it seems like there will be continuity where it's needed unless I'm missing something.

That works for Atom, but what about other people? I'm very, very reticent to ship features that you need to know a "special trick" in order to not get broken

paulcbetts:
![](http://cl.ly/image/0S2l322J3v3Y/Atom_-_Slack_2015-09-28_14-44-08.png)

I have no idea how many places in the code rely on being able to use the prerelease tag for delta vs full, which is why this change is terrifying. You'll for sure have to update `ReleaseEntry`, then do basically an audit of the code looking for other places that care about delta vs full, or try to parse the release file name in any way.

paulcbetts:
I told this to @maxbrunsfeld, but I think there is actually a better way to do beta channels / releases that doesn't involve the `-beta` tag (The current proposal involves massively changing Squirrel and very possibly breaking / stranding users, without any user-visible benefit).

The better way to do this is (we'll say here we have a 'beta' and 'stable' release channel):

1. Every build of Atom that gets released gets a real version number (i.e. 1.1.2)
1. The Beta version of the RELEASES file gets every version number
1. Certain releases get Promoted to the Stable channel - this involves simply copying the nupkg files over and adding a line to Stable's RELEASED entry

This means you can promote releases to Stable without making a new version number or build, you just copy some files. It also means that version numbers are monotonically increasing across channels, which is easier to understand.

nathansobo:
I really prefer the self-describing nature of our current plan. It is also monotonic, and additionally you can immediately determine from the version number alone what stage of the release pipeline the release is associated with and what version it will graduate to in the next phase of the pipeline. I don't want to have to consult a table to determine if a version is a beta.

thedaniel:
[...]

Issue #93, opened 2014-10-14
Possible issue with version numbers being used in install directory


GeertvanHorrik:
When I install this nuget version:

MyProduct-1.7.0-unstable0036-full.nupkg

That has NuGet version 1.7.0-unstable0036, it is installed as:

app-1.7.0

That menas that I can never update to a new version because it will also pick 1.7.0. Why is the -unstable0036 removed from the path?

joaomoreno:
I would like to see this happen.

DanPristupov:
Same here. I'd like to use prerelease numbers in Squirrel too.

GeertvanHorrik:
Can you explain why you closed this one?

paulcbetts:
@GeertvanHorrik Mostly to discourage the Me Toos

joaomoreno:
That just discourages the support of the project.

PR #450, opened 2015-09-29
Allow prerelease versions when parsing release entries

maxbrunsfeld:
Fixes https://github.com/Squirrel/Squirrel.Windows/issues/449
Fixes https://github.com/Squirrel/Squirrel.Windows/issues/93

@paulcbetts  I'm opening this now so that you can :gun: it down sooner rather than later.

#### Overview

My strategy here is to avoid changing the format of `RELEASES`. The `-delta` and `-full` suffixes are still used to denote delta vs full packages, but semver pre-release strings are *also* allowed.

Unlike @GeertvanHorrik's proposed solution, I'm just relying on `NuGet.SemanticVersion` and its built-in ordering relation (prerelease versions are sorted alphabetically). No hard-coded orderings of specific prerelease strings.

#### Status

* AFAICT, there are twenty tests failing on `master`, and the same twenty fail against this branch.
* I haven't done any manual testing yet. I'll test some scenarios using Atom if you think this approach is viable.

### DIFF
--- a/src/Squirrel/ReleaseExtensions.cs
+++ b/src/Squirrel/ReleaseExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using NuGet;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -9,39 +10,19 @@ namespace Squirrel
 {
     public static class VersionExtensions
     {
-        public static Version ToVersion(this IReleasePackage package)
+        private static readonly Regex _suffixRegex = new Regex(@"(-full|-delta)?\.nupkg$", RegexOptions.Compiled);
+        private static readonly Regex _versionRegex = new Regex(@"\d+(\.\d+){0,3}(-[a-z][0-9a-z-]*)?$", RegexOptions.Compiled);
+
+        public static SemanticVersion ToSemanticVersion(this IReleasePackage package)
         {
-            return package.InputPackageFile.ToVersion();
+            return package.InputPackageFile.ToSemanticVersion();
         }
 
-        public static Version ToVersion(this string fileName)
+        public static SemanticVersion ToSemanticVersion(this string fileName)
         {
-            var parts = (new FileInfo(fileName)).Name
-                .Replace(".nupkg", "").Replace("-delta", "")
-                .Split('.', '-').Reverse();
-
-            var numberRegex = new Regex(@"^\d+$");
-
-            var versionFields = parts
-                .Where(x => numberRegex.IsMatch(x))
-                .Select(Int32.Parse)
-                .Reverse()
-                .ToArray();
-
-            if (versionFields.Length < 2 || versionFields.Length > 4) {
-                return null;
-            }
-
-            switch (versionFields.Length) {
-            case 2:
-                return new Version(versionFields[0], versionFields[1]);
-            case 3:
-                return new Version(versionFields[0], versionFields[1], versionFields[2]);
-            case 4:
-                return new Version(versionFields[0], versionFields[1], versionFields[2], versionFields[3]);
-            }
-
-            return null;
+            var name = _suffixRegex.Replace(fileName, "");
+            var version = _versionRegex.Match(name).Value;
+            return new SemanticVersion(version);
         }
     }
 }
--- a/src/Squirrel/ReleaseExtensions.cs
+++ b/src/Squirrel/ReleaseExtensions.cs
@@ -10,8 +10,8 @@ namespace Squirrel
 {
     public static class VersionExtensions
     {
-        private static readonly Regex _suffixRegex = new Regex(@"(-full|-delta)?\.nupkg$", RegexOptions.Compiled);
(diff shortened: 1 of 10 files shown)


src/Squirrel/ReleaseExtensions.cs
```diff
@@ -9,39 +10,19 @@ namespace Squirrel
 {
     public static class VersionExtensions
     {
-        public static Version ToVersion(this IReleasePackage package)
+        private static readonly Regex _suffixRegex = new Regex(@"(-full|-delta)?\.nupkg$", RegexOptions.Compiled);
```
paulcbetts: You don't usually need the `private` keyword in C#, the default visibility is "as private as possible"

maxbrunsfeld: Noted. I was copying `NuGet`'s style here with these static `Regex` instances.

GeertvanHorrik:
Good to hear someone is picking this up again, well done. However... The string replacements were there for a reason. 

In your case, it means 1.0-unstable123 is higher than 1.0-beta1. Is that really what you want? How are you handling unstable, alpha and beta? I think stuff like "unstable" and "nightly" should never be considered higher than alpha or beta. And since alpha is the first letter, it's hard to find something that comes in before that that makes sense to end users.

From what I remember, NuGet's logic also is case sensitive, so "1.0.0-unstable15" is lower than "1.0.0-Unstable14". So please be very careful when using something that is written for NuGet...

Note that all situations were covered with unit tests in my PR as well, so not sure why Paul didn't accept it. Hopefully yours will be accepted.

And now we are at it. I also added a feature (to my custom branch) that adds the release date. This is *very* useful in lots of scenarios (showing how long a release has been out and the user is still not on that release, or for license checking scenarios). It's a simple (non-breaking) addition since the parser can just assume null when a release date is unknown (old behavior).

maxbrunsfeld:
> In your case, it means 1.0-unstable123 is higher than 1.0-beta1

Yeah, I can't see any reason that `unstable123` should be lower than `beta1`. From the specification on semver.org (emphasis mine):

> Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows: **identifiers consisting of only digits are compared numerically and identifiers with letters or hyphens are compared lexically in ASCII sort order**.

Here is what `NuGet.SemanticVersion` does to compare pre-release strings:

```cs
return StringComparer.OrdinalIgnoreCase.Compare(SpecialVersion, other.SpecialVersion);
```

Admittedly, they aren't exactly the same. NuGet's logic is case-*insensitive*, whereas semver stipulates ASCII-order (capitals first). Also, NuGet requires that pre-release names start with letters and don't contain any dots.

Overall, I think that matching NuGet's behavior is an appropriate way for Squirrel to deal with pre-releases.

paulcbetts:
> Note that all situations were covered with unit tests in my PR as well, so not sure why Paul didn't accept it. Hopefully yours will be accepted.

I'm just as apprehensive now as I was then but this is a blocking bug for Atom's beta program so we're going to try it and see who gets broken also I still hate this

GeertvanHorrik:
Then I am still wondering how atom will handle nightlies and betas. What names will you be using?

nathansobo:
If we ever do nightlies, they'll always be for the next version. So if the latest beta is `1.2.0-beta-0`, then the latest nightly will be `1.3.0-dev`. This version states that that nightly will eventually be incorporated into what becomes Atom 1.3.0.

GeertvanHorrik:
Yes, and how are you going to release 1.3.0-beta-0? Anyway, I solved this by using alpha as nightly builds.

nathansobo:
> Yes, and how are you going to release 1.3.0-beta-1

If I understand your question correctly, as soon as `1.3.0-beta-1` was released, nightlies would start being `1.4.0-dev-SHA`, indicating the final stable release version that would contain the code in the nightly.

GeertvanHorrik:
But according to semver, 1.3.0-beta-1 is *smaller* than 1.3.0-unstable-1 or 1.3.0-dev-1

Haacked:
> But according to semver, 1.3.0-beta-1 is smaller than 1.3.0-unstable-1 or 1.3.0-dev-1

That is correct. If you need metadata in the version that doesn't affect precedence, you need to use the build metadata hash `#`. For example `1.3.0-beta-1#unstable`.

Haacked:
> Admittedly, they aren't exactly the same. NuGet's logic is case-insensitive, whereas semver stipulates ASCII-order (capitals first). Also, NuGet requires that pre-release names start with letters and don't contain any dots.
> 
> Overall though, I think that matching NuGet's behavior is an appropriate way for Squirrel to deal with pre-releases.

Keep in mind that there's an [open PR](https://github.com/mojombo/semver/pull/177) to change the comparison of labels to be case insensitive.

I'm inclined to accept this PR but the person who submitted is no longer on the NuGet team and I've been very busy with other stuff.

paulcbetts:
Thanks @maxbrunsfeld!

maxbrunsfeld:
:zap: Thanks for throwing me a :meat_on_bone: on this.

paulcbetts:
I'll put out a release on Monday or so

Issue #439, opened 2015-09-12
Improve "Getting started" page


Matthias-Hess:
Just an idea how to improve the Getting Started documentation to make it a bit more "fool proof". I see two possible problems with the following code snippet:

```
// NB: For this version, always say your app is using .NET 4.5, even if it's
// totally not
using (var mgr = new UpdateManager("https://path/to/my/update/folder")) 
{
    await mgr.UpdateApp();
}
```

### first problem
If you just copy and paste the code snippet into your application, the code will not compile, because UpdateManager is in a namespace that is not in scope.

The novice has two possible ways to resolve this:

- he adds a using directive at the top of the file. *If he does this. everything is fine.*
- he uses the fully qualified name. *If he chooses this option, he might get confused, because UpdateManager does not have an UpdateApp() method, as this is an extension method.*

It would be good to include a hint like *add the following using directive at the top of your code*:

```
using Squirrel;
```

### second problem
The comment in the code snippet about .NET 4.5 seems completeley out of context. You can safely remove that comment.

elbaloo:
@Ribeiro did u forget to submit ur PR?

Ribeiro:
@elbaloo Actually I already did but it seems to be ignored https://github.com/Squirrel/Squirrel.Windows/pull/458

Ribeiro:
@elbaloo or @paulcbetts  could you please close the issue? Thankx  =)

elbaloo:
@Ribeiro [this](https://help.github.com/articles/closing-issues-via-commit-messages/) will come handy for your next PR ;)

Ribeiro:
@elbaloo Thankx ;)

PR #458, opened 2015-10-06
Fixed issue #439 related to Getting started documentation page

Ribeiro:
Please, check pull request. Thanx  =)


paulcbetts:
I think the original comment is more important, we don't need to teach the basics of C# in our READMEs. Thanks though!

paulcbetts:
Actually, that comment in the old version is complete nonsense!

PR #459, opened 2015-10-07
Handle query strings in target URLs

### LINKED ISSUE #132: Invalid RELEASES url if query params are in the original url
If you use a URL like `https://localhost:3000/updates?version=0.147` to `--download` then the `RELEASES` file will be requested at `https://localhost:3000/updates?version=0.147/RELEASES` instead of `https://localhost:3000/updates/RELEASES?version=0.147`

maxbrunsfeld:
Fixes #132 

This allows the following commands to handle URLs with query parameters
* `SyncReleases.exe --url=<URL>`
* `Update.exe --download=<URL>`
* `Update.exe --update=<URL>`

#### Testing
I temporarily published a beta release of Atom so that the `atom.io` redirects to that release's `RELEASES` asset when it gets a request for `/api/updates/RELEASES?version=1.1.0-beta0`.

* [x] Run `SyncReleases.exe --url=https://atom.io/api/updates?version=1.1.0-beta0` and verify that the correct `RELEASES` file is used.

  * Before: Fails with `System.Net.Http.HttpRequestException: Response status code does not indicate success: 404`
  * After: Works

* [x] Run `Update.exe --download=https://atom.io/api/updates?version=1.1.0-beta0` and verify that the correct release entries are returned.

  * Before: Fails with `System.Exception: Remote release File is empty or corrupted`
  * After: Works

* [x] Run `Update.exe --update` and verify that the app updates to the correct version.

  * Before: Fails with `System.Exception: Remote release File is empty or corrupted`
  * After: Updates from Atom 1.0.19 to 1.1.0-beta1

### DIFF
--- a/src/SyncReleases/SyncImplementations.cs
+++ b/src/SyncReleases/SyncImplementations.cs
@@ -14,7 +14,7 @@ namespace SyncReleases
     {
         public static async Task SyncRemoteReleases(Uri targetUri, DirectoryInfo releasesDir)
         {
-            var releasesUri = new Uri(targetUri + "/RELEASES");
+            var releasesUri = Utility.AppendPathToUri(targetUri, "RELEASES");
             var releasesIndex = await retryAsync(3, () => downloadReleasesIndex(releasesUri));
             File.WriteAllText(Path.Combine(releasesDir.FullName, "RELEASES"), releasesIndex);
 
@@ -24,7 +24,7 @@ namespace SyncReleases
                 .Take(5)
                 .Select(x => new {
                     LocalPath = Path.Combine(releasesDir.FullName, x.Filename),
-                    RemoteUrl = new Uri(targetUri + "/" + x.Filename)
+                    RemoteUrl = new Uri(Utility.EnsureTrailingSlash(targetUri), x.BaseUrl + x.Filename)
                  });
 
             foreach (var releaseToDownload in releasesToDownload) {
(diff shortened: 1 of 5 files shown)


src/SyncReleases/SyncImplementations.cs
```diff
@@ -24,7 +24,7 @@ internal class SyncImplementations
                 .Take(5)
                 .Select(x => new {
                     LocalPath = Path.Combine(releasesDir.FullName, x.Filename),
-                    RemoteUrl = new Uri(targetUri + "/" + x.Filename)
+                    RemoteUrl = new Uri(Utility.EnsureTrailingSlash(targetUri), x.BaseUrl + x.Filename)
```
maxbrunsfeld: The addition of `BaseUrl` here is somewhat unrelated to this change, but since I was touching this line already, it seemed right to include that in the URL, like it is in [`UpdateManager.downloadRelease`](https://github.com/Squirrel/Squirrel.Windows/blob/b9f804b24024ae701777d18e2539693aa51328ca/src/Squirrel/UpdateManager.DownloadReleases.cs#L71). Let me know what you think of that.

maxbrunsfeld:
Ok @paulcbetts I think this is ready to go; I've tested it for Atom.

paulcbetts:
Fixed a few very minor syntax nits, but :+1:!  You've got a promising career as a C# developer @maxbrunsfeld

maxbrunsfeld:
Thanks @paulcbetts. :beers:

PR #464, opened 2015-10-12
Extend GitHubUpdateManager to allow an accessToken to avoid GitHub API rate limiting

### LINKED ISSUE #463: GitHubUpdateManager should allow access tokens to avoid the GitHub API rate limit
Should just need to update [this line](https://github.com/Squirrel/Squirrel.Windows/blob/master/src/Squirrel/UpdateManager.Factory.cs#L48) to put the path and query in the right place. If a token is supplied, it should generate a URL in this form:

```
https://api.github.com/repos/<user>/<repo>/releases?access_token=<token>
```

stajs:
Fixes #463.

### DIFF
--- a/src/Squirrel/UpdateManager.Factory.cs
+++ b/src/Squirrel/UpdateManager.Factory.cs
@@ -34,7 +34,8 @@ namespace Squirrel
             string applicationName = null,
             string rootDirectory = null,
             IFileDownloader urlDownloader = null,
-            bool prerelease = false)
+            bool prerelease = false,
+            string accessToken = null)
         {
             var repoUri = new Uri(repoUrl);
             var userAgent = new ProductInfoHeaderValue("Squirrel", Assembly.GetExecutingAssembly().GetName().Version.ToString());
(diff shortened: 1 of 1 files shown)


BarryThePenguin:
Looks good.

I assumed 1 minute per IP would be suitable when checking for updates, sorry I missed that

stajs:
No problem, easy to miss. Until you hit the dreaded :skull: _Enterprise_ :skull: scenario with hundreds of people sharing an external IP! :boom:

src/Squirrel/UpdateManager.Factory.cs
```diff
@@ -43,9 +44,16 @@ public class Release
                 throw new Exception("Repo URL must be to the root URL of the repo e.g. https://github.com/myuser/myrepo");
             }
 
+            var releasesApiBuilder = new StringBuilder("/repos")
+                .Append(repoUri.AbsolutePath)
```
stajs: I've carried on using the path directly here. It might be better to explicitly ask for username and repo parameters. This could still blow up if we end up getting an unexpected path here.

paulcbetts:
Sorry I missed this @stajs, this'll go into the next release. Thanks!

stajs:
Groovtatstic :+1:

paulcbetts:
@stajs Due to me being bad at computer, this is already out as 1.1.2

PR #466, opened 2015-10-14
Create IT Administrator-friendly MSI files

paulcbetts:
### What is this?

This PR adds an extra step to Releasify that creates an MSI based off of our installer, suitable for IT Administrators to deploy in their environment via Group Policy. 

### Hey, this MSI doesn't work!?

This MSI isn't a general-purpose installer, it is designed to mimic the behavior of `--machine` (#337 for the lazy) - this means that once you run the MSI, users from now on will get the app installed, *on Login*. So, most normal users should continue to run the Setup.exe's generated by Releasify, but if you want to have an IT Admin Friendly version, you can hand off the MSI


MiroljubA:
I have tried both Setup.exe --machine and Setup.msi however still none will install machine wide.  Setup.exe --machine installs for current user and Setup.msi doens't install at all, only adds an entry to Installed apps named "Machine wide install".
Windows event log doesn't  show anything. Any ideas? Thanks !

MiroljubA:
Solved. The title attribute in the nuspec file was empty.

hypnotisis:
Hi

I'm not sure I understand the purpose of the MSI. For some reasons (on specific machines), I need to have a single application installed (for example in the "Program Files" folder). Does the MSI allow me to do this? Or will it install the application for all users by installing as much as user?

giemch:
Where is the install location for these msi's?

Neustradamus:
All good softwares when we install use by default "Program Files" (or recently "Program Files (x86)" for 32 bit apps in 64 bit system).
Bad softwares do not use it.

It is since Windows 95 and maybe before, no?

anaisbetts:
@Neustradamus Spamming three separate issues with your uninformed opinion is inappropriate. If you continue you will be banned from this repo.

Neustradamus:
@anaisbetts: Thanks for your reply :)
What do you think about the issue?

lukeskinner:
@Neustradamus Anyone can fork the code on GitHub with those changes or submit a PR (provided it doesn't break backwards compatibility for existing users of this library). 

What's the issue with how it is at the moment as you haven't explained why it's causing you a problem? The project owners of Squirrel are very careful about not breaking existing users of this library by changing small things.

PR #469, opened 2015-10-16
Fallback to exe icon if iconUrl empty

radiy:
In many cases it is impossible to know where setup will be hosted, so it is impossible to know icon url beforehand
Often setup contains single exe file so is reasonable to set uninstall icon to exe icon


paulcbetts:
It seems weird to pick a random EXE out of the air to use as the icon

PR #471, opened 2015-10-26
GitHub enterprise support

CSharpFan:
* Squirrel now has ability to fetch releases from an Enterprise version of GitHub (b7eebcd80e192b6be26c2e3a7dc0e9ab68ea7887)
* Fixed the issue that when you said 'yes, I want prerelease versions', it would exclude normal release versions with a higher version number (7304c747fb4d0f6af763046fb87959302a69fe6c). 
* Using the `.Length()` method on a `List<T>` instead of `.Count()` extension method. Former is faster. (ae63e7ae565c996e31099c87176f0728c3d83698)


kid:
+1 on this

BarryThePenguin:
I like this a lot :+1:

paulcbetts:
@CSharpFan Thanks!

PR #499, opened 2015-11-21
Allow access_token from repoUrl

KevinMartin:
Adds the ability to specify access token in repo url. This is useful in case you want to do something like:

```
Update.exe --download https://github.com/atom/atom?access_token=abc123
```


KevinMartin:
I realized after I created this PR that `Update.exe` doesn't utilize the GitHubUpdateManager.

So, I implemented my own update server.

PR #519, opened 2015-12-11
Don't copy the Mono variant of Update.exe into embedded zip

fasterthanlime:
cf. https://github.com/atom/grunt-electron-installer/pull/80#issuecomment-156272858

Note that the generated Setup.exe still has trouble for me ([here's the log](https://gist.github.com/fasterthanlime/86407596c8f813ca910c)), am a bit lost on what to do next


fasterthanlime:
Opened #520 to deal with further issues

paulcbetts:
:+1: :sparkles:

PR #522, opened 2015-12-12
--signTool support

fasterthanlime:
This is based on #505.

Instead of automatically selecting Mono's signtool, it lets the user choose which sign tool to use via the `--signTool` command line option.

If `osslsigncode` is specified, the executable to sign is copied to a temporary file, and the temporary file is appended to the argument list, as osslsigncode needs its input and output files to be different.

I'm not fond of having osslsigncode-specific tool, but as far as I can tell it's the only viable option for doing Authenticode on non-Windows platforms

@paulcbetts since you asked in #505, there *might* be a way to `SignTool.exe` work with wine, in the future, with a lot of effort, but I'm not holding my breath:

  * <http://projects.zillabit.com/authenticode.html> (Ctrl-F "failed attempts to")
  * <https://www.winehq.org/pipermail/wine-users/2007-March/026599.html>

FWIW, I don't consider mono's "signtool" a viable alternative either: there's unavoidable `.bak` files (could be deleted from Update-Mono.exe with, again, tool-specific codepaths), and it has been printing "Succeeded" yet producing files with invalid signatures.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -87,6 +87,7 @@ namespace Squirrel.Update
                 string bootstrapperExe = default(string);
                 string backgroundGif = default(string);
                 string signingParameters = default(string);
+                string signTool = default(string);
                 string baseUrl = default(string);
                 string processStart = default(string);
                 string processStartArgs = default(string);
@@ -120,7 +121,8 @@ namespace Squirrel.Update
                     { "g=|loadingGif=", "Path to an animated GIF to be displayed during installation", v => backgroundGif = v},
                     { "i=|icon", "Path to an ICO file that will be used for icon shortcuts", v => icon = v},
                     { "setupIcon=", "Path to an ICO file that will be used for the Setup executable's icon", v => setupIcon = v},
-                    { "n=|signWithParams=", "Sign the installer via SignTool.exe with the parameters given", v => signingParameters = v},
+                    { "signTool=", "Path to code signing utility (defaults to SignTool.exe)", v => signTool = v},
+                    { "n=|signWithParams=", "Sign the installer via SignTool.exe or other with the parameters given", v => signingParameters = v},
                     { "s|silent", "Silent install", _ => silentInstall = true},
                     { "b=|baseUrl=", "Provides a base URL to prefix the RELEASES file packages with", v => baseUrl = v, true},
                     { "a=|process-start-args=", "Arguments that will be used when starting executable", v => processStartArgs = v, true},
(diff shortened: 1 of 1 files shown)


src/Update/Program.cs
```diff
@@ -570,22 +572,44 @@ void waitForParentToExit()
...
+                }
+            }
 
-                // Run down PATH and hope for the best
-                if (!File.Exists(exe)) exe = "signtool.exe";
+            String tmpExe = null;
+
+            if (signTool == "osslsigncode") {
+                // osslsigncode needs input != output, and to have them 
+                tmpExe = exePath + ".unsigned";
```
fasterthanlime: might be worth generating a unique name or even a temp directory? was going to say I didn't know how to do that in C# but actually this very same file contains an example of that.

src/Update/Program.cs
```diff
@@ -570,22 +572,44 @@ void waitForParentToExit()
...
+                File.Copy(exePath, tmpExe);
+                signingOpts = signingOpts + " " + tmpExe;
             }
 
+            // SignTool.exe requires `sign` parameter (it can also verify, etc.)
+            // signcode (mono) ignores `sign` parameter
+            // osslsigncode ignores `sign` parameter
```
fasterthanlime: thought it was worth noting.

src/Update/Program.cs
```diff
@@ -570,22 +572,44 @@ void waitForParentToExit()
...
+                    // Run down PATH and hope for the best
+                    if (!File.Exists(exe)) exe = "signtool.exe";
+                }
+            }
 
-                // Run down PATH and hope for the best
-                if (!File.Exists(exe)) exe = "signtool.exe";
+            String tmpExe = null;
+
+            if (signTool == "osslsigncode") {
```
fasterthanlime: I feel pretty bad about comparing magic strings (twice), if there's a will to add better support for mono's signcode too, it's probably worth refactoring a bit, to something less imperative perhaps.

halfbyte:
Just to give another boost to this PR (which, I know, needs some work): I'm currently trying to set up automatic building and signing at a company where the cert comes from a smartcard and the whole signing is done with a python script that calls a build job on the sever with the smartcard.

This may sound ridiculous at first, but sharing hardware certs between different build jobs that by necessity have to run on different machines is a valid concern, I think.

Is there anything specific (apart from currently having conflicts, obviously) that keeps this from getting merged? It's been lying around for a while now?

fasterthanlime:
I have to mention I'm not using Squirrel.Windows actively anymore (nor am I making windows builds on non-windows platforms). I won't be able to follow up on this PR, but the code is there and you should feel free to pick it up!

halfbyte:
@fasterthanlime Thanks for the heads up. Will try to do, if I get any sort of positive signal that this has any chance of getting merged ;)

PR #536, opened 2015-12-29
Enhanced the troubleshooting and getting-started to prevent some pitfalls

Lakritzator:


docs/getting-started.md
```diff
@@ -14,7 +14,7 @@ using (var mgr = new UpdateManager("https://path/to/my/update/folder"))
 }
 ```
 
-* Use NuGet Package Explorer (or any other way) to create a NuGet package for your app. Make sure that package doesn't have any dependencies. Here's a good example package:
+* Use NuGet Package Explorer (or any other way) to create a NuGet package (.nupkg) from a .nuspec for your app. Make sure that package doesn't have any dependencies. Here's a good example package:
```
JVimes: I don't think .nuspec is necessarily required if using NuGet Package Explorer, though it can be used.

PR #549, opened 2016-01-12
Reformatting of docs (fix)

kenbailey:
Caught a link that got missed and an inconsistency in the previous commits.


paulcbetts:
Thanks!

PR #551, opened 2016-01-12
Fixed path to nuget-package-metadata from install-process

JakeGinnivan:


JakeGinnivan:
Ugh.. github ui doesn't respect line endings?

JakeGinnivan:
Will do it properly

PR #553, opened 2016-01-13
Fixed some path issues and added naming info

JakeGinnivan:


kenbailey:
Thanks @JakeGinnivan, I'm including these updates into a doc update PR I will be submitting and hopefully get them rolled in.

Issue #508, opened 2015-12-01
MSI based installation fails


knutt:
We are trying to use the msi file generated by Squirrel. The msi file seems to run ok without error, and installs a file named \<ApplicationPackageName\>.exe in folder "c:\program files(x86)\\\<ApplicationName> Installer". When run, this file displays the following error message:
"This is a dummy setup.exe. If you are seeing this, your build did not correctly replace the IDR_UPDATE_ZIP resource.
Command line: '--install . ' "

So, my question is: How do I correctly replace the IDR_UPDATE_ZIP resource?

paulcbetts:
Does your normal Setup.exe run without error?

knutt:
Yes, the setup.exe built by Squirrel is fine. No problem there.

paulcbetts:
Huh, really? That is quite odd. Can you share the log file that gets generated when you run --releasify?

knutt:
Please find the log file attached. (Renamed the file to ...txt to allow it to be attached)
[SquirrelSetup.log.txt](https://github.com/Squirrel/Squirrel.Windows/files/50546/SquirrelSetup.log.txt)

skaryshev:
I checked out 1.2.1 and got the same issue. The problem is template.wxs points to stub (bootstrapper) setup.exe in the same directory. I patched my template.wxs with hardcoded path to target setup.exe and the issue gone.

paulcbetts:
@skaryshev Sure, but the template is copied to releases where the Setup.exe is supposed to be already filled out

skaryshev:
Right, but I think  ".\Setup.exe" path in template.wxs->setup.wsx is relative to candle.exe which is in squirrel.exe working directory.

knutt:
Thanks,

I changed the setup.exe path in template.wsx to: Source="../../../../Release/Setup.exe". Problem solved for now.

paulcbetts:
@skaryshev Can you make a PR to this effect?

skaryshev:
Cant push local branch to github:
`git push origin skaryshev_msi_fix`
`remote: Permission to squirrel/squirrel.windows.git denied to skaryshev`

paulcbetts:
@skaryshev In this case, you need to hit the fork button and push to your fork of Squirrel.Windows, then send a PR from the website - here's a good guide: http://stackoverflow.com/questions/14680711/how-to-do-a-github-pull-request

PR #556, opened 2016-01-13
Proposed MSI build fix

skaryshev:
Proposed MSI build fix to make createMsiPackage(...) explicitly supply path to releases directory to both candle and light. Target setup.exe is located in releases directory. This fix prevents MSI builder from picking up incorrect (bootstrapper) setup.exe located in working directory (tools).

### DIFF
--- a/src/Squirrel/Utility.cs
+++ b/src/Squirrel/Utility.cs
@@ -177,7 +177,7 @@ namespace Squirrel
             }
         }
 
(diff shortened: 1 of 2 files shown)


skaryshev:
The issue was initially discussed in https://github.com/Squirrel/Squirrel.Windows/issues/508

paulcbetts:
@skaryshev Thanks! And thanks for putting up with the GitHub nonsense.

PR #558, opened 2016-01-14
InvokeProcessAsync workingDirectory param default to empty string

skaryshev:
According to Microsoft https://msdn.microsoft.com/en-us/library/system.diagnostics.processstartinfo.workingdirectory%28v=vs.110%29.aspx ProcessStartInfo.WorkingDirectory Property default value is empty string. It does not affect anything but in such sensitive area better be safe than sorry.


paulcbetts:
:+1:

PR #564, opened 2016-01-19
Problem exists in CE 2015 as well

StillLearnin:


kenbailey:
Thanks @StillLearnin.

Issue #566, opened 2016-01-21
Mono.Cecil question


skaryshev:
One and the only place we currently using Mono.Cecil API is
`SquirrelAwareExecutableDetector.GetAssemblySquirrelAwareVersion`
This method may be easy re-implemented using regular System.Reflection calls. (I tried it and the tests passed with new implementation).
If we are not planning wider utilization of Mono.Cecil API in the future, we can eliminate this dependency entirely.

paulcbetts:
The problem with trying to use Reflection is that you have to load the assembly, which means that it will try to load all of its children into your process, which has been Highly Problematic in the past (the original version of Squirrel would implement hooks as C# classes, and it was a disaster)

skaryshev:
Well, to my surprise, according to MSDN https://msdn.microsoft.com/en-us/library/ms172331%28v=vs.110%29.aspx there is a way to load assembly without dependencies. It sounds like `Assembly.ReflectionOnlyLoadFrom`  loads assembly not into AppDomain but to isolated reflection-only context. Do you think it worth to try?

paulcbetts:
@skaryshev Mayyyybe? We can try it I guess.

skaryshev:
I am trying 'Assembly.ReflectionOnlyLoadFrom' now. One of the tests is failing
`SquirrelAwareExecutableDetectorTests.SquirrelAwareViaAssemblyAttribute' that one where the test is probing its own assembly for Squirrel awareness. I guess 'Assembly.ReflectionOnlyLoadFrom' does not allow loading assembly which is already in AppDomain. The rest of the tests are passing. How important for us is this particular test case? If it is important we can query AppDomain first to get already loaded assembly.

skaryshev:
PR submitted for further discussion https://github.com/Squirrel/Squirrel.Windows/pull/567

PR #567, opened 2016-01-22
Get SquirrelAwareVersion without Mono.Cecil

skaryshev:
To continue discussion https://github.com/Squirrel/Squirrel.Windows/issues/566
Mono.Cecil is way too powerful to use it in only one place. On the other hand regular System.Reflection calls require loading of all assembly dependencies and not suitable to retrieve only one attribute from one assembly. They also have a nasty side effects in execution of static members initialization on loading application into AppDomain.
Proposed solution is built around `Assembly.ReflectionOnlyLoadFrom`


skaryshev:
Found a big problem here. When we are probing assembly A decorated with
`[assembly: MyCoolAssemblyAttribute()]` and type `MyCoolAssemblyAttribute` in defined in assembly B then assembly B must be loaded which defeats whole idea of not loading dependencies. I am about to close this PR.

paulcbetts:
@skaryshev It was a good try!

skaryshev:
Without Mono.Cecil we are better off just scanning assembly file for a string literal "SquirrelAwareVersion" or even better for some pattern containing "SquirrelAwareVersion" but this is entirely different story.

PR #575, opened 2016-01-28
Update nuget-package-metadata.md

provkristian:
Hello. I thought that it might be useful to emphasize this so that folks don't lose time on something like this. Thanks.


kenbailey:
@provkristian thanks!

provkristian:
Thanks @kenbailey !

Issue #358, opened 2015-05-25
Error when installing when is already install


ricardopolo:
Hello.

Did you have seen this behavior:
1. You install an app
2. You uninstall it
3. You try to install it again and it fails

It's because after you uninstall it the app is still running. (Related to #318)

So you have any plan for support it, or should we manage in the Uninstall event?

ricardopolo:
@paulcbetts any comments?

paulcbetts:
@ricardopolo Squirrel can't know how your application works so it can't know what to kill - you should shell out to `TaskKill.exe /F /IM Image.exe /T` in your uninstall hook

PR #590, opened 2016-02-05
Kill processes belonging to us on install and uninstall

paulcbetts:
For certain types of apps, it's super easy to run the installer while the app is still running. This of course fails, but users think it is just because the installer has a bug. Enact Process Violence to stop this from occurring


src/Squirrel/NativeMethods.cs
```diff
@@ -7,8 +8,30 @@
...
 {
-    static class NativeMethods
+    public static class NativeMethods
     {
+        public static int GetParentProcessId()
+        {
+            var pbi = new PROCESS_BASIC_INFORMATION();
+
+            //Get a handle to our own process
+            IntPtr hProc = OpenProcess((ProcessAccess)0x001F0FFF, false, Process.GetCurrentProcess().Id);
```
CharlieHess: Can you use the enum here instead of the literal?

paulcbetts: This is just moved code from long ago, but yeah, probably

CharlieHess:
Could maybe beef up the tests around this scenario, although there's no way to mock them.

ricardopolo:
@paulcbetts so this is not necesary anymore?
https://github.com/Squirrel/Squirrel.Windows/issues/358

paulcbetts:
@ricardopolo Nope!

PR #601, opened 2016-02-15
Make splash/loading window draggable

damieng:
It's a little annoying that the splash/loading screen might be obscuring something (mine was) even for just 10 seconds... but making it draggable is easy.


paulcbetts:
:clap:

PR #605, opened 2016-02-18
Fix Content Types created by Mono

### LINKED ISSUE #520: Update-Mono generates invalid [Content-Types].xml
`[Content_Types].xml`

```xml
﻿<?xml version="1.0" encoding="utf-8"?>
<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
  <Default ContentType="application/vnd.openxmlformats-package.relationships+xml" Extension="rels" />
  <Default ContentType="application/octet" Extension="nuspec" />
  <Default ContentType="application/octet" Extension="bin" />
  <Default ContentType="application/octet" Extension="dll" />
  <Default ContentType="application/octet" Extension="pak" />
  <Default ContentType="application/octet" Extension="exe" />
  <Default ContentType="application/octet" Extension="dat" />
  <Default ContentType="application/octet" Extension="" />
  <Default ContentType="application/vnd.openxmlformats-package.core-properties+xml" Extension="psmdcp" />
  <Default Extension="diff" ContentType="application/octet" />
  <Default Extension="bsdiff" ContentType="application/octet" />
  <Default Extension="shasum" ContentType="text/plain" />
</Types>
```

See [this gist](https://gist.github.com/fasterthanlime/86407596c8f813ca910c#file-squirrellog-txt-L81) for full stack trace, it's the System.Xml package complaining, while Nuget is trying to ensure we have a valid manifest.

I have no idea how it's generated, if someone knows, all help is appreciated, otherwise I'll probably keep digging tomorrow.

(My instinct is that this line (the one after 'dat') should probably just not get generated at all?)

KevinMartin:
Attempts to remove default content types in `[Content_Types].xml` with empty Extensions.

This happens when adding a file via your NuSpec file with no extension, such as `LICENSE`.

Fixes #520 

/cc @fasterthanlime

### DIFF
--- a/src/Squirrel/ContentType.cs
+++ b/src/Squirrel/ContentType.cs
@@ -6,6 +6,23 @@ namespace Squirrel
 {
     internal static class ContentType
     {
+        public static void Clean(XmlDocument doc)
+        {
+            var typesElement = doc.FirstChild.NextSibling;
+            if (typesElement.Name.ToLowerInvariant() != "types") {
+                throw new Exception("Invalid ContentTypes file, expected root node should be 'Types'");
+            }
+
+            var children = typesElement.ChildNodes.OfType<XmlElement>();
+            for (var child in children)
+            {
+                if (child.GetAttribute("Extension") == "")
+                {
+                    typesElement.RemoveChild(child);
+                }
+            }
+        }
+
         public static void Merge(XmlDocument doc)
         {
             var elements = new [] {
--- a/src/Squirrel/ReleasePackage.cs
+++ b/src/Squirrel/ReleasePackage.cs
@@ -349,6 +349,7 @@ namespace Squirrel
             doc.Load(path);
 
             ContentType.Merge(doc);
+            ContentType.Clean(doc);
 
             using (var sw = new StreamWriter(path, false, Encoding.UTF8)) {
                 doc.Save(sw);


fasterthanlime:
This is basically a workaround for a nupkg bug right? It should probably be fixed upstream instead.

KevinMartin:
I'm not really familiar with nupkg/nuget, unfortunately.

Plus, this is backwards compatible with previous versions of nupkg/nuget.

paulcbetts:
Yeah, while I agree with @fasterthanlime in principle, in reality this is probably going to be Very Arduous

KevinMartin:
I also agree. The ideal scenario would be to fix upstream. But...

- The [NuGet.exe in Squirrel.Windows](https://github.com/Squirrel/Squirrel.Windows/blob/master/.nuget/NuGet.exe) is 2 years old.
- And the [nuget.exe in atom/grunt-electron-installer](https://github.com/atom/grunt-electron-installer/blob/master/vendor/nuget.exe) is 1 year old.
- And the `nuget.exe` in all of the popular forks of `atom/grunt-electron-installer`, such as `electronjs/windows-installer`, `Aluxian/electron-windows-installer`, and `mongodb-js/electron-installer-squirrel-windows` all use the same 1 year old `nuget.exe`.

fasterthanlime:
I was going to mention that the workaround would be a good idea if it was one of the last holdouts for getting a complete Squirrel.Windows toolchain working on Linux.. but the Microsoft patching package is the real problem there, not the LICENSE file.

However, if nobody bothers updating their NuGet.exe, then that workaround might be very welcome/useful. :+1:

KevinMartin:
@paulcbetts is this good to be merged then or are there any changes you would like me to do?

paulcbetts:
@KevinMartin It's probably fine, I'll look at it soon

paulcbetts:
Thanks!

PR #614, opened 2016-02-26
Remove machine install as a Thing

paulcbetts:
This PR removes machine installation now that we have MSI packages, and more importantly hard-blocks folx from running our installer as an elevated administrator.

This PR also improves the MSI `checkInstall` code to completely refuse to do anything if the package directory exists in any way. This prevents installers from getting into this "rolling install" scenario where we somehow decide to pave over our own installation, then continually corrupt it instead.

## TODO:

- [x] Test with normal UAC settings
- [x] Test with UAC disabled

### DIFF
--- a/src/Setup/MachineInstaller.cpp
+++ b/src/Setup/MachineInstaller.cpp
@@ -4,51 +4,6 @@
 #include "resource.h"
 #include <sddl.h>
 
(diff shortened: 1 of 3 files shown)


CharlieHess:
:+1:

PR #616, opened 2016-03-01
Use months instead of minutes for InstallDate

### LINKED ISSUE #613: Funny small issue with install date
I think you set the minute instead of the month in the installed date. (in registry)
Because of these sometimes the install date has an invalid format.

khellang:
Closes #613


paulcbetts:
:+1:

Issue #641, opened 2016-03-25
UpdateManager.GitHubUpdateManager Not Updateing


dartvalince:
I have tried multiple way of using the GitHubUpdateManager to update the application, and both have not worked at all.   The first method I used was the one straight from the example documentation at 

https://github.com/Squirrel/Squirrel.Windows/blob/master/docs/using/github.md

When using that, this was the App.xaml.cs code

    public partial class App : Application
    {
        protected override async void OnStartup(StartupEventArgs e)
        {
            base.OnStartup(e);

                 using (var mgr = UpdateManager.GitHubUpdateManager("https://github.com/dartvalince/DiscerningEye/"))
                 {
                     await mgr.Result.UpdateApp();
                 }
            
            
            //  Check for updats to the alarm data file
            this.UpdateAlarmData();

            var bs = new Bootstrapper();
            bs.Run();

            DiscerningEye.App.Current.Exit += Current_Exit;

            DiscerningEye.Properties.Settings.Default.PropertyChanged += Default_PropertyChanged;

            //  Apply startup theme from users saved settings
            ChangeApplicationTheme();

        }

When using this i went through all the steps in the documentation

* Commit
* Create .nupkg
* releasify
* Create new release on GItHub adding all the files from the `Releases` folder

After this, I wen through the use case a user would and downloaded the files from the GitHub release and ran the Setup.exe. It installed the application correctly, placing items in the Start Menu, on the Desktop, and in the local %appdata% folder, however, the application did not start.  I checked the processes running in Task Manager, and it showed the process running as a Background Process, with 0% CPU and 0% Network and ~7MB of memory.

I figured the problem with this was that the `await` command was not releasing control back to the `OnStartup` method.  So I then adjusted the App.xaml.cs to the following 

    public partial class App : Application
    {
        protected override void OnStartup(StartupEventArgs e)
        {

            base.OnStartup(e);

            Task.Run(async () =>
             {
                 using (var mgr = UpdateManager.GitHubUpdateManager("https://github.com/dartvalince/DiscerningEye/"))
                 {
                     await mgr.Result.UpdateApp();
                 }
             });
            
            
            //  Check for updats to the alarm data file
            this.UpdateAlarmData();

            var bs = new Bootstrapper();
            bs.Run();

            DiscerningEye.App.Current.Exit += Current_Exit;

            DiscerningEye.Properties.Settings.Default.PropertyChanged += Default_PropertyChanged;

[...]

Issue #611, opened 2016-02-19
Can the setup.exe / setup.msi names be changed ? 


btoda:
Can we have MyProductSetup.exe ?

paulcbetts:
Yes! You can just rename Setup.exe yourself.

btoda:
Thanks for the quick reply.

Issue #574, opened 2016-01-28
How do I check if the app is installed via Squirrel?


JonasDralle:
I want to distribute my app as Squirrel edition and Standalone edition (the standalone wont be able to update)

How can I find out weather an app is installed via Squirrel or not?
My current method of determining isn't very good especially becaue you guys just used System.Execption instead of something like NonSquirrelInstallationException

            UpdateManager updateManager;

            try
            {
                updateManager = new UpdateManager("https://company.com/software/releases");
            }
            catch (Exception exe)
            {
                if (exe.Message.Contains("Update.exe not found"))
                {
                    //This isn't an squirrel installed app
                    return;
                }

                //throw everything else
                throw exe;
            }

JonasDralle:
If you don't have a good solution I would like it if you could implement one

paulcbetts:
```cs
var assembly = Assembly.GetEntryAssembly();
var updateDotExe = Path.Combine(Path.GetDirectoryName(assembly.Location), '..', 'Update.exe');
var isInstalled = File.Exists(updateDotExe);
```

JonasDralle:
oh thank you

quasarea:
It looks like this does the job:

    var updates = await updateManager.CheckForUpdate();
    if (updates.CurrentlyInstalledVersion == null) { 
      //Seems like this is not a squirrel installation, running within VisualStudio perhaps?
      MessageBox.Show("Unknown current version - possibly application was not installed correctly");
    } else if (updates.ReleasesToApply.Any()) {
      //Squirrel installation with updates to apply
      await updateManager.DownloadReleases(updates.ReleasesToApply);
      await updateManager.ApplyReleases(updates);
      await updateManager.UpdateApp();
    }

GregWilliamBryant:
> It looks like this does the job:
> 
> ```
> var updates = await updateManager.CheckForUpdate();
> if (updates.CurrentlyInstalledVersion == null) { 
>   //Seems like this is not a squirrel installation, running within VisualStudio perhaps?
>   MessageBox.Show("Unknown current version - possibly application was not installed correctly");
> } else if (updates.ReleasesToApply.Any()) {
>   //Squirrel installation with updates to apply
>   await updateManager.DownloadReleases(updates.ReleasesToApply);
>   await updateManager.ApplyReleases(updates);
>   await updateManager.UpdateApp();
> }
> ```

[...]

Issue #218, opened 2015-02-06
Kaspersky Anti-virus identifies Setup.exe and Update.exe as trojans


chrispret:
Kaspersky's heuristic scan detects both Setup.exe and Update.exe as HEUR:Trojan.Win32.Generic.

Obviously it isn't liking some of the squirrel code. While I was happy to scan the files using a few different engines and none of them resulted in detections, typical end-users might not agree and end up not installing the software.

When I build from the source code, the resulting Setup.exe is detected as well.

It was also a pain to download and install Squirrel.windows through the Visual Studio Package Manager, since Kaspersky deleted Setup.exe every time it was downloaded. I eventually figured out what was going on and set it to prompt me before deleting the file, which made it possible to install it.

paulcbetts:
While this was originally a problem, these days as long as you sign your installer, most people aren't seeing the virus warnings. Can you send in Setup.exe and Update.exe to Kapersky to whitelist then?

chrispret:
I'll submit them.

Is the setup.exe that is distributed through the Visual Studio Package manager signed? If so, there might be value in submitting new versions to them when publishing. If I didn't realize that my anti-virus was deleting the files (without even telling me) I would have never gotten the package installed on my project.

paulcbetts:
> Is the setup.exe that is distributed through the Visual Studio Package manager signed?

No, because the setup.exe we distribute there is a template that we'll edit (and therefore, break the signature)

cwoolum:
I had the same issue with Bit defender but I sent in Setup.exe and they added it to their database. Not sure if that is possible with Kapersky as well

JVimes:
Symantec Endpoint Protection's generic trojan detector, [Trojan.Gen.2](http://securityresponse.symantec.com/security_response/writeup.jsp?docid=2011-082216-3542-99), quarantined test\fixtures\Setup.exe. I submitted it on their [false positive form](https://submit.symantec.com/false_positive/).

paulcbetts:
Thanks @JVimes

jg123:
Windows Defender quarantined the unsigned Update.exe as a Trojan (Trojan:Win32/Codinx.B!plock). I submitted it as not malware here: https://www.microsoft.com/security/portal/submission/SubmissionHistory.aspx?SubmissionId=a1235358-fbb7-4d39-83a8-942440050036&lc=1033

sawilde:
[...]

PR #642, opened 2016-03-26
Doc Updates (#218, #574, #611,  #641)

kenbailey:
FAQ updates based on responses to #218, #574, and #611.
Trailing slash comment from #641.


kenbailey:
Just a few updates @paulcbetts

paulcbetts:
:+1:

Issue #603, opened 2016-02-16
Spaces in file name turn into %20


fdreher:
I have a file called `unity default resources` in my project's build. After installing via the generated setup.exe, the file is unpacked as `unity%20default%20resources` and my application does not work.
NuGet Package Explorer correctly shows the file name with spaces when browsing the *.nupkg, so I suspect that the issue is related to Squirrel.

Unfortunately, renaming the file in my project is not an option because I cannot control this particular file name.

geary:
Duplicate of #414 and #544.

Until this is fixed in Squirrel, you can use [this workaround](https://github.com/Squirrel/Squirrel.Windows/issues/544#issuecomment-172672313) I posted a few weeks ago. It simply renames the file to the correct spelling.

fdreher:
How did I manage to miss these two issues while googling the problem? Sorry for cluttering up the issue tracker.

raybooysen:
This is a dupe of #544

Issue #544, opened 2016-01-06
Squirrel replaces spaces in filename with %20


Deozaan:
A dependency I have no control over has spaces in one of the filenames. The NuGet Package manager translates the spaces to %20 in the nuget package. After releasifying the package in Squirrel and using Setup.exe to install, things are broken because the file that should have spaces in its name has %20 in there instead.

I tried editing the package file to change the %20 to spaces, but that just broke things earlier in the process (Squirrel wouldn't even package it up into a setup.exe file).

To be clearer: It is NuGet's fault that the spaces get turned into %20 when creating the package. But Squirrel should be smart enough to convert the %20 to a space when it is unpacking/creating the files.

geary:
I've run into this too. Any Unity app will have this problem because of this file: `appname_Data\Resources\unity default resources`

I wonder if this just affects spaces or other "escapable" characters too?

Looking through the Squirrel code, the one bit of code that seems related is the `Uri.EscapeUriString()` call in `compressFolderEncoded()` and the `Uri.UnescapeDataString()` call in `extractZipDecoded()` (both in `src\Squirrel\ReleasePackage.cs`).

At first it seemed obvious that these were matched up properly, but if NuGet is doing *another* escape that would be a different story. Maybe Squirrel needs to do another unescape, or not do this escape?

In the meantime, since it's just that one file breaking my app, I worked around it by adding this bit of code to my launcher app (a "Squirrel aware" app that runs before the Unity app):

    static void FixupFiles() {
        FixupSpaces(
            Path.Combine( DataDir, "Resources" ),
            "unity default resources"
        );
    }

    static void FixupSpaces( string dir, string file ) {
        var newFile = Path.Combine( dir, file );
        var oldFile = Path.Combine( dir, file.Replace( " ", "%20" ) );
        if( File.Exists( oldFile ) ) {
            File.Delete( newFile );
            File.Move( oldFile, newFile );
        }
    }

And then I call `FixupFiles()` in the `onInitialInstall` and `onAppUpdate` events.

`DataDir` is a static property with the name of the Unity data directory, e.g. `MyAppMain_Data`. Adjust as needed for a non-Unity app.

Deozaan:
Yeah, Unity is the dependency I mentioned. Thanks for posting the workaround.

That said, I hope Squirrel fixes this soon.

Ran-Intendu:
Duplicates #414

Thieum:
[...]

Issue #414, opened 2015-08-25
File names are URI escaped.


Ran-Intendu:
"File Name" is installed as "File%20Name"

paulcbetts:
Which file names?

Ran-Intendu:
A file which is included in the nuget package we create.
Specifically, I have a file called "unity default resources" which is being extracted as "unity%20default%20resources".

paulcbetts:
Probably related to https://github.com/Squirrel/Squirrel.Windows/pull/394/files

duanyao:
Nuget expects entry names in .nupkg file %-escaped, so if you found "unity%20default%20resources" in the .nupkg file or extract it with a zip tool, it is normal. After installaton with Squirrel, the file names should be unescaped (If not, it must be a bug).

Ran-Intendu:
Well, it is a bug.

When I open the nuget package through 7zip, I can see that the file is %-encoded.
When I open the nuget package through NuGet Package Explorer, the file is decoded (without %).
Finally, after installation with Squirrel, the file remains escaped --> Bug.

paulcbetts:
I think this is a combination of @duanyao's change and my change that rips NuGet out of the extraction code.

hfcorriez:
+1

hankbao:
Same here.

hankbao:
Hi @paulcbetts. Do you have any ETA for the fix of this issue?

jbeshir:
Got bit by the same problem here. Now I know about it I can at least change things to avoid spaces in names as a workaround. Interestingly enough doesn't seem to happen for @ symbols, just spaces. Maybe it just doesn't consider that symbol as something needing escaping.

ruisebastiao:
+1

yuanyinchun:
+1

geary:
I posted some code to [work around this](https://github.com/Squirrel/Squirrel.Windows/issues/544#issuecomment-172672313) in the other issue a few weeks ago. It may be helpful for anyone who runs into this.

kumarharsh:
I am also running into the same bug, but it's because I've named my own app with spaces in the name.  The app exe now as a `%20` in it's name. Everything works correctly, but it does look ugly !

Thieum:
@shiftkey fixed by #650 , this can be closed.

bddckr:
[...]

PR #650, opened 2016-04-05
Fix: File names URI unescape (aka %20). #414 #544 #603.

Ran-Intendu:
Please review it thoroughly.


paulcbetts:
@Ran-Intendu Thanks for this!

MarshallOfSound:
This has caused MASSIVE regression problems

* In electron calling `updateAndRestart()` doesn't restart the app as the `.exe` path has changed
* All previous shortcuts to the app no longer work as the `.exe` path has changed

This basically renders all installs of an application useless as there is no easy user friendly way to launch the application any more

PR #658, opened 2016-04-08
MSI consistent hashes for UpgradeCode

paulcbetts:
Right now, we currently generate random GUIDs for our MSI component IDs. This means, that if you use GPOs to push down an MSI file for v1.0, then push down another MSI file for v2.0, we get the same app installed twice / weird stuff happens / etc etc. Instead, we'll base our GUIDs on a derivative of the package ID


CharlieHess:
:shipit:

PR #663, opened 2016-04-14
Redundant kill all processes

### LINKED ISSUE #662: Uninstall one SquirrelAware CefSharp app, others crash.
When creating a SquirrelAware app, I noticed when I uninstalled it GitHub for Windows would crash.

Testing this with multiple SquirrelAware apps, I can crash separate applications by uninstall other SquirrelAware apps.

I think it comes down to [this here](https://github.com/Squirrel/Squirrel.Windows/blob/8674e343fa30b7970a40a80b98eedec37330a5f3/src/Squirrel/UpdateManager.ApplyReleases.cs#L128).

``` Csharp
// NB: Some people attempt to uninstall apps while 
// they're still running. I cannot even.
var toKill = allApps
    .SelectMany(x => Process.GetProcessesByName(x.Name.Replace(".exe", "")))
    .ToList();
```

and more specifically [this](https://github.com/Squirrel/Squirrel.Windows/blob/9de260cb6b614234511d57723738f1be2ed25476/src/Squirrel/SquirrelAwareExecutableDetector.cs#L16)
``` CSharp
var di = new DirectoryInfo(directory);

return di.EnumerateFiles()
    .Where(x => x.Name.EndsWith(".exe", StringComparison.OrdinalIgnoreCase))
    .Select(x => x.FullName)
    .Where(x => (GetPESquirrelAwareVersion(x) ?? -1) >= minimumVersion)
    .ToList();
```

I could be reading this wrong, but basically it's looking for all `exe` files in the app install directory and nuking any (and all) running instances of any found exe files, matching by name. This causes a problem when multiple application are going to be using `CefSharp.BrowserSubprocess.exe` for example.

Shouldn't this process only care for the main executable that is being uninstalled rather than killing any and all running instances matching a name of an exe in the install directory?

Any help is appreciated! :+1:

Layoric:
Fix to resolve #662, seems the problem code is actually redundant due to a call to `KillAllExecutablesBelongingToPackage` in `UpdateManager.FullUninstall` prior to `ApplyReleasesImpl.FullUninstall` being called.

Now GitHub for Windows doesn't crash! :tada:

### DIFF
--- a/src/Squirrel/UpdateManager.ApplyReleases.cs
+++ b/src/Squirrel/UpdateManager.ApplyReleases.cs
@@ -122,17 +122,6 @@ namespace Squirrel
                         } else {
                             allApps.ForEach(x => RemoveShortcutsForExecutable(x.Name, ShortcutLocation.StartMenu | ShortcutLocation.Desktop));
                         }
-
-                        // NB: Some people attempt to uninstall apps while 
-                        // they're still running. I cannot even.
-                        var toKill = allApps
-                            .SelectMany(x => Process.GetProcessesByName(x.Name.Replace(".exe", "")))
-                            .ToList();
-
-                        if (toKill.Count > 0) {
-                            toKill.ForEach(x => x.Kill());
-                            Thread.Sleep(750);
-                        }
                     } catch (Exception ex) {
                         this.Log().WarnException("Failed to run pre-uninstall hooks, uninstalling anyways", ex);
                     }


paulcbetts:
:+1:

Layoric:
@paulcbetts Thanks! Out of interest, any plans on when the next NuGet update will be?

paulcbetts:
Sometime either on Friday or early next week

PR #666, opened 2016-04-19
Staged rollouts

paulcbetts:
This PR implements staged rollouts, where a certain percentage of your users will get the newest version of your app.

We do this by adding a magic comment to the RELEASES file - unfortunately, because of a bug in the syntax Regex in existing Squirrel.Windows versions, we have to write this in a super ugly way:

```
A3D8B15ABDB5F2A3D9FEBDB8F55B6D4FB8168828 atom-1.9.0-dev-4e74fb6-full.nupkg 110216564# 10%
```

Note that there is _no_ space between the filesize and the comment. In the future, you'll able to put a space there of course :)

## TODO:

- [x] Write some tests and what
- [x] Verify RELEASES format against old version of Squirrel.Windows
- [x] Verify case where we're not in the beta percentage
- [x] Rig beta ID, see us get put into the beta

### DIFF
--- a/src/Squirrel/ReleaseEntry.cs
+++ b/src/Squirrel/ReleaseEntry.cs
@@ -88,16 +88,23 @@ namespace Squirrel
 
         static readonly Regex entryRegex = new Regex(@"^([0-9a-fA-F]{40})\s+(\S+)\s+(\d+)[\r]*$");
         static readonly Regex commentRegex = new Regex(@"#.*$");
+        static readonly Regex stagingRegex = new Regex(@"#\s+(\d{1,3})%$");
         public static ReleaseEntry ParseReleaseEntry(string entry)
         {
             Contract.Requires(entry != null);
+
+            float? stagingPercentage = null;
+            var m = stagingRegex.Match(entry);
+            if (m != null) {
+                stagingPercentage = Single.Parse(m.Groups[0].Value);
+            }
 
             entry = commentRegex.Replace(entry, "");
             if (String.IsNullOrWhiteSpace(entry)) {
                 return null;
             }
 
-            var m = entryRegex.Match(entry);
+            m = entryRegex.Match(entry);
             if (!m.Success) {
                 throw new Exception("Invalid release entry: " + entry);
             }
@@ -138,7 +145,7 @@ namespace Squirrel
             long size = Int64.Parse(m.Groups[3].Value);
             bool isDelta = filenameIsDeltaFile(filename);
 
-            return new ReleaseEntry(m.Groups[1].Value, filename, size, isDelta, baseUrl, query);
+            return new ReleaseEntry(m.Groups[1].Value, filename, size, isDelta, baseUrl, query, stagingPercentage);
         }
 
         public static IEnumerable<ReleaseEntry> ParseReleaseFile(string fileContents)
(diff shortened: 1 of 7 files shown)


src/Squirrel/ReleaseEntry.cs
```diff
@@ -130,7 +145,18 @@ public static ReleaseEntry ParseReleaseEntry(string entry)
             long size = Int64.Parse(m.Groups[3].Value);
             bool isDelta = filenameIsDeltaFile(filename);
 
-            return new ReleaseEntry(m.Groups[1].Value, filename, size, isDelta, baseUrl, query);
+            return new ReleaseEntry(m.Groups[1].Value, filename, size, isDelta, baseUrl, query, stagingPercentage);
+        }
+
+        public bool IsStagingMatch(Guid? userId)
```
CharlieHess: Some comments on how we define a staging match would be 👌.

src/Squirrel/UpdateManager.CheckForUpdates.cs
```diff
@@ -177,6 +178,37 @@ UpdateInfo determineUpdateInfo(IEnumerable<ReleaseEntry> localReleases, IEnumera
...
+
+                try {
+                    if (!Guid.TryParse(File.ReadAllText(stagedUserIdFile, Encoding.UTF8), out ret)) {
+                        throw new Exception("File was read but contents were invalid");
+                    }
+
+                    this.Log().Info("Using existing staging user ID: {0}", ret.ToString());
+                    return ret;
+                } catch (Exception ex) {
+                    this.Log().InfoException("Couldn't read staging user ID, creating a blank one", ex);
```
CharlieHess: Probably don't need to write this exception out since it'll always happen the first time.

test/ReleaseEntryTests.cs
```diff
@@ -88,6 +88,26 @@ public void ParseVersionTest(string releaseEntry, int major, int minor, int patc
             Assert.Equal(isDelta, fixture.IsDelta);
         }
 
+        [Theory]
```
CharlieHess: `Theory`? `InlineData`? What is this wizardry?

KevinMartin:
@paulcbetts Maybe add Docs to that Todo?

kenbailey:
@paulcbetts if you can point me to some example code, I'll can create a doc file for you to review.

test/ReleaseEntryTests.cs
paulcbetts: Parameterized unit tests!

src/Squirrel/ReleaseEntry.cs
```diff
@@ -122,15 +137,29 @@ public static ReleaseEntry ParseReleaseEntry(string entry)
...
+        {
+            // A "Staging match" is when a user falls into the affirmative
+            // bucket - i.e. if the staging is at 10%, this user is the one out
+            // of ten case.
+            if (!StagingPercentage.HasValue) return true;
+            if (!userId.HasValue) return false;
+
+            uint val = BitConverter.ToUInt32(userId.Value.ToByteArray(), 12);
+
+            double percentage = ((double)val / (double)UInt32.MaxValue);
```
stajs: It seems to me, that in the absence of tracking actual downloads/installs, it is just assuming an even distribution of guid-to-integer-conversions? So, a hopeful approximation?

paulcbetts: We create the GUID in a Weird Way by taking random bytes that is more likely to ensure a random distribution, it's not just `Guid.Create()`. That being said, I am not a statistician, but it doesn't _really_ matter so much

src/Squirrel/UpdateManager.CheckForUpdates.cs
```diff
@@ -177,6 +178,37 @@ UpdateInfo determineUpdateInfo(IEnumerable<ReleaseEntry> localReleases, IEnumera
 
                 return UpdateInfo.Create(currentRelease, remoteReleases, packageDirectory);
             }
+
+            internal Guid? getOrCreateStagedUserId()
+            {
+                var stagedUserIdFile = Path.Combine(rootAppDirectory, "packages", ".betaId");
```
stajs: I couldn't tell from the docs how the rollout determination is happening. It would be nice if the docs gave some background on this `.betaId` file.

paulcbetts: Hm, so users / developers shouldn't have to know this implementation detail, but I don't see why we can't add it to the docs. It's effectively just a random number, serialized in GUID format

stajs: For the most part, you are probably right, but for my 2 cents, I wanted to know at least the high-level view of how it worked before I jumped in to trying it out.

PR #668, opened 2016-04-20
Added newline for better rendering

kysel:


kenbailey:
Thanks @kysel, we could use a merge when you get a chance @paulcbetts.

PR #677, opened 2016-04-30
Update the .nuspec dependencies

taspeotis:
Hi,

Squirrel seems to be a major user of my DeltaCompressionDotNet library. The current version is v1.1. I have a v2 I'd like to release with some small-but-breaking changes.

Squirrel's `.nuspec` depends on v1 *or higher* (see [here](https://docs.nuget.org/create/versioning) for more information).

This pull request changes the `.nuspec` to accept only v1-based versions of the library.

I'm not in any rush to release v2. So if you do accept this change I'm content to wait until Squirrel releases an updated NuGet package in advance of my own NuGet package.

Thanks.


paulcbetts:
Thanks!

Issue #545, opened 2016-01-06
Setup.exe doesn't support version numbers with more than 2 digits


Deozaan:
I like to put the date in my version numbers as an easy way to know how old a release is. But that leads to strange behavior in the installer created by Squirrel.

Somehow my version 1.141122.10 is being changed to 1.66.10.0. And if I try to split it up a little by using version 1.14.1122.10 then somehow that becomes 1.14.98.10.

![squirrelversionweirdness](https://cloud.githubusercontent.com/assets/227734/12139009/6cfe1880-b41a-11e5-9b93-8ea7a3a1e5d5.png)

Additionally (and this perhaps should be submitted as a separate issue), it seems the "Product Name" is taken from the NuGet Description rather than the NuGet Title.

paulcbetts:
The Windows PE file format limits all of those fields to a single byte, you have to have a number < 256

geary:
These version fields are actually unsigned 16-bit values, not 8-bit values. So `1.141122.1` is not a valid version value (the max value for any field is `65535`), but `1.14.1122.10` is perfectly valid and should be allowed.

The real issue here is that rcedit is erroneously truncating these values to 8 bits.

I fixed the problem in rcedit, tested it with Squirrel, and put up a [PR here](https://github.com/atom/rcedit/pull/7).

If you grab the updated version of rcedit it will fix this Squirrel issue.

PR #682, opened 2016-05-06
Update rcedit.exe

bontibon:
The updated binary is taken from: https://github.com/atom/node-rcedit/tree/master/bin.

It includes the fix for #545 (via atom/rcedit#7).


paulcbetts:
Thanks!

PR #716, opened 2016-05-29
Print update info stdout

mbsanchez:
I have added a new command to the update tool, now non c# apps, can check for update, receiving the same information as download command, but this new command doesn't download the release, only check for it.

update commands

Also, I have logged some update message to stdout, so, the parent app can check the message and get some information about the update status using a pipe.

Please consider adding this to the main branch.


paulcbetts:
Thanks!

PR #725, opened 2016-06-05
Used package title for product name.

### LINKED ISSUE #724: Request: Consider using package "title" as the product name
Currently setup file uses for the product name the package `description` or `summary` or `id`.
Consider switching to package `title` and fallback to `id`. Usage of `description` or `summary` doesn't reflect the purpose of the field.

JimiC:
Resolves #724.


paulcbetts:
This would break existing software for no real user benefit, sorry!

JimiC:
I strongly suggest to reconsider this. Never the less, you're the owner and I respect your decision.

paulcbetts:
I would probably agree that if we were designing it from scratch we'd use Title but changing it now will cause problems with Squirrel users

PR #735, opened 2016-06-10
Provide a way to specify custom application exit logic for UpdateManager.RestartApp()

szanto90balazs:
I want to be able to specify how the application exits when ```UpdaterManager.RestartApp()``` is called.

This is the default implementation:

```
exeToStart = exeToStart ?? Path.GetFileName(Assembly.GetEntryAssembly().Location);
var argsArg = arguments != null ? String.Format("-a \"{0}\"", arguments) : "";
exiting = true;

Process.Start(getUpdateExe(), String.Format("--processStartAndWait {0} {1}", exeToStart, argsArg));

Thread.Sleep(500);
Environment.Exit(0);
```

The only part I want to make customizable is:
```
Environment.Exit(0);
```

My biggest problem with ```Environment.Exit(0);``` that it has stopped working with our multi-process application. I tried to make it work, but it failed and even if I could fix it, I'm 100% sure that the code would be scattered around in the code base in a form of additional ```if```s and magic :sparkles: ```bool```:sparkles:s. Maintenance pain.

I also suspect, that ```Environment.Exit(0);``` doesn't allow the application to gracefully terminate, that's why I would like to call ```Application.Current.Shutdown(0)```, which allows a graceful termination in a form of firing the ```OnExit``` WPF event.

As a quick workaround I copy pasted the ```UpdateManager.RestartApp``` and the ```UpdateManager.getUpdateExe``` code, but as ```getUpdateExe``` is private ...

### DIFF
--- a/src/Squirrel/UpdateManager.cs
+++ b/src/Squirrel/UpdateManager.cs
@@ -173,35 +173,23 @@ namespace Squirrel
         }
 
         static bool exiting = false;
-        public static void RestartApp(string exeToStart = null, string arguments = null)
-        { 
-            // NB: Here's how this method works:
-            //
-            // 1. We're going to pass the *name* of our EXE and the params to 
-            //    Update.exe
-            // 2. Update.exe is going to grab our PID (via getting its parent), 
-            //    then wait for us to exit.
-            // 3. We exit cleanly, dropping any single-instance mutexes or 
-            //    whatever.
-            // 4. Update.exe unblocks, then we launch the app again, possibly 
-            //    launching a different version than we started with (this is why
-            //    we take the app's *name* rather than a full path)
 
-            exeToStart = exeToStart ?? Path.GetFileName(Assembly.GetEntryAssembly().Location);
-            var argsArg = arguments != null ?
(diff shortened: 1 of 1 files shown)


paulcbetts:
Why don't you clean up what you're doing *then* call RestartApp?

szanto90balazs:
Finally, I managed to find out what should I do to make all process terminate (namely ```CefRuntime.Shutdown()``` on the main thread), so Squirrel wouldn't wait forever to start the application again.

Now my workflow is:

1. When it comes to restarting the application, I call my ```AppFinalizer.OnAppExit``` to clean things up 
2. As a last step I call ```CefRuntime.Shutdown()``` and ```UpdateManager.RestartApp``` so ```UpdateManager.RestartApp``` can succeed

Now it works as expected and we don't need this PR. I only favored the idea of calling ```Application.Current.Shutdown(0)```, because it just worked out of the box and it might be a slightly more bullet proof solution and less code.

paulcbetts:
The problem is that not all apps respond to Application.Current, only WPF ones

paulcbetts:
Thanks for contributing the PR though! :)

PR #746, opened 2016-06-21
Use entry exe path instead of Squirrel.dll when looking for Update.exe

engdata:
Currently, the documentation states that "The UpdateManager is expecting to find the Update.exe application installed one directory up from the EXE"; however this is currently not true. It is looking for Update.exe one directory above the Squirrel.dll, which may be commonly in an assemblies subdirectory. Using Assembly.GetEntryAssembly (with a fallback to the executing assembly in case of unmanaged code) allows for proper functioning.

### DIFF
--- a/src/Squirrel/UpdateManager.cs
+++ b/src/Squirrel/UpdateManager.cs
@@ -279,7 +279,7 @@ namespace Squirrel
                 return Path.GetFullPath(assembly.Location);
             }
 
-            assembly = Assembly.GetExecutingAssembly();
+            assembly = Assembly.GetEntryAssembly() ?? Assembly.GetExecutingAssembly();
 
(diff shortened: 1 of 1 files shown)


paulcbetts:
I'm into it. Is this the only place we do this?

engdata:
Hi Paul - I have searched and walked through the code and this appears to be the only instance where this is an issue. There are other uses of 'GetExecutingAssembly' however all appear to be acceptable uses (like checking against the root path or investigating the squirrel.dll assembly itself). 
This is definitely the only place where the assembly path is used to find the update.exe program. I believe this change should likely have been made at the same time as [this commit](https://github.com/Squirrel/Squirrel.Windows/commit/2ccc6a7d9bb8fd8047ee1badaeccd8c92c80d29e), which correctly changed the first usage to use GetEntryAssembly.

paulcbetts:
:+1:. thanks!

PR #758, opened 2016-07-04
Fix --processStart issues with <= 2.2.0

MarshallOfSound:
Replace %20 with spaces in versions above 2.2.0 so when launching shortcuts created in <= 2.2.0 we can still find the exe.

Fixes electron/windows-installer/issues/78

/cc @paulcbetts


paulcbetts:
👍

PR #763, opened 2016-07-14
Patch to address the issue of moving directories with Antivirus Scan

albertsuchvi:
This pull request is a 'patch' to address the problem seen when installing on a system with antivirus scan.

**The problem**
As part of the installation process, the .nupkg is unpacked in a temp directory, and then some of the directories under that package are moved to the final place where they have to be installed.
When the system has an antivirus scan, sometimes that move fails, because some of teh files in the directories that have to be moved may be being scanned by the scanner, and that causes the move to fail because the files are in use.

**The "solution"**
The change makes the install process resilient to this situation, by trying several times the move before issuing an error. 
I know that the fix is not very elegant (sorry, I am not an expert in c# programming), but it has worked on the cases we had problems installing the application that we are developping

### DIFF
--- a/src/Squirrel/UpdateManager.ApplyReleases.cs
+++ b/src/Squirrel/UpdateManager.ApplyReleases.cs
@@ -32,6 +32,8 @@ namespace Squirrel
             public async Task<string> ApplyReleases(UpdateInfo updateInfo, bool silentInstall, bool attemptingFullInstall, Action<int> progress = null)
             {
                 progress = progress ?? (_ => { });
+                
+                this.Log().Info("Starting to apply releases");
 
                 var release = await createFullPackagesFromDeltas(updateInfo.ReleasesToApply, updateInfo.CurrentlyInstalledVersion);
                 progress(10);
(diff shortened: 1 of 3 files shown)


src/Squirrel/UpdateManager.ApplyReleases.cs
```diff
@@ -302,18 +306,62 @@ Task<string> installPackageToAppDir(UpdateInfo updateInfo, ReleaseEntry release)
...
                     toMove.ForEach(ld => {
+                        this.Log().Info("Move item " + ld.FullName + "to " + target.FullName);
+            
                         ld.GetDirectories()
-                            .ForEachAsync(subdir => subdir.MoveTo(subdir.FullName.Replace(ld.FullName, target.FullName)))
-                            .Wait();
+                           .ForEachAsync(subdir => {
+                               this.Log().Info("Move subdir " + subdir.FullName + " to " + subdir.FullName.Replace(ld.FullName, target.FullName));                            
+                               Boolean moved = false;
+                               Int16 trials = 0;
```
paulcbetts: There is a method called `Utility.Retry` that does this pretty well already, no need to reinvent the wheel!

paulcbetts:
Hey @albertsuchvi, thanks for having a hack at solving this problem! I decided to fix it a better way (i.e. by not doing the move in the first place) in #764

Issue #769, opened 2016-07-20
SyncRemoteReleases does not pass along query parameters when requesting files


mbutters:
While building an application using a populated releases file that contains urls with query string parameters, once the files are requested they do not contain the query parameters.  Example:

Release File contents:

`F137844DD5C32B49752131B05ZXXED0328 https://somedomain.com/somepath/files/filename-1.0.0-full.nupkg?release=5C9272399XA5DRD8369670`

Squirrel will request: https://somedomain.com/somepath/files/filename-1.0.0-full.nupkg (excluding the parameters)

The cause of this issue is: https://github.com/Squirrel/Squirrel.Windows/blob/master/src/SyncReleases/SyncImplementations.cs#L28

The URL is manually being built with the domain and file name and not including the query parameters.  The fix would be to also include the query parameters.

mbutters:
Pull request has been created for this: https://github.com/Squirrel/Squirrel.Windows/pull/770

Andrew-Hanlon:
Can I ask how you setup your nuspec/squirrel to get the domain name in the Releases file? This (along with your fix) would allow hosting on S3 with 'private' urls.

mbutters:
Our releases file is remotely hosted and is generated once requested by our internal file list that we store in a database.  The domain name either can be generated on the fly (as the current domain) or can be placed statically.

mbutters:
Also to add to this, it looks like the UpdateManager does pass along the query parameters:  https://github.com/Squirrel/Squirrel.Windows/blob/9de260cb6b614234511d57723738f1be2ed25476/src/Squirrel/UpdateManager.DownloadReleases.cs#L70  but the SyncImplementation does not.  We were wondering why there was a inconsistency between them and figured it was an oversight.

Andrew-Hanlon:
Ah, interesting thanks for the additional info. So in fact I can already set this up just by manually adding the query parts to my Releases. Going to try it.

Thieum:
@shiftkey PR fixing this issue has been merged, this can be closed.

PR #770, opened 2016-07-20
Update SyncImplementations.cs

mbutters:
Update for SyncImplementations.cs to pass along query string parameters when requesting files. More info can be viewed at https://github.com/Squirrel/Squirrel.Windows/issues/769

### DIFF
--- a/src/SyncReleases/SyncImplementations.cs
+++ b/src/SyncReleases/SyncImplementations.cs
@@ -25,7 +25,7 @@ namespace SyncReleases
                 .Take(1)
                 .Select(x => new {
                     LocalPath = Path.Combine(releasesDir.FullName, x.Filename),
-                    RemoteUrl = new Uri(Utility.EnsureTrailingSlash(targetUri), x.BaseUrl + x.Filename)
+                    RemoteUrl = new Uri(Utility.EnsureTrailingSlash(targetUri), x.BaseUrl + x.Filename + x.Query)
                  });
 
             foreach (var releaseToDownload in releasesToDownload) {


paulcbetts:
I'm into this as a concept, I'm just not sure it's correct

mbutters:
Which part did you think is not correct?  The code change or something else?  FYI we did compile it on our end with the change and it did seem to work fine.  Our testing wasn't extensive but we were able to pass along query parameters and fetch what we were looking for.

paulcbetts:
@mbutters How does this code build a correctly formatted URL just by concatenating?

mbutters:
Although it is simple, the concatenation was following a similar pattern found here: 
https://github.com/Squirrel/Squirrel.Windows/blob/9de260cb6b614234511d57723738f1be2ed25476/src/Squirrel/UpdateManager.DownloadReleases.cs#L71

The query attribute is either [populated or null](https://github.com/Squirrel/Squirrel.Windows/blob/1aebe10db3e9d9b3d5a4c147622299d3a6073a30/src/Squirrel/ReleaseEntry.cs#L137).  In the case that it is null, C# will output it as an empty string.  When it is populated the query attribute of the Release Entry is pulled from the [Uri.Query](https://msdn.microsoft.com/en-us/library/system.uri.query(v=vs.110).aspx) attribute which will pull in the "?" and any key value pairs.  There are a few other ways of achieving the same results, however since the base url and file name were already being concatenated it seemed to make sense to also concatenate the query string to minimize the changes.

PR #778, opened 2016-07-26
Patch appUserModelID so that it doesn't contain spaces as per spec

### LINKED ISSUE #773: appUserModelId can not contain a space
As per the spec here --> https://msdn.microsoft.com/en-us/library/windows/desktop/dd378459(v=vs.85).aspx#which

> An application must provide its AppUserModelID in the following form. It can have no more than 128 characters and cannot contain spaces. Each section should be camel-cased.

https://github.com/Squirrel/Squirrel.Windows/blob/c699425abc0f808de764de874b46cbdb7195db4c/src/Squirrel/UpdateManager.ApplyReleases.cs#L185

Needs to be changed to remove spaces.  And a follow up PR needs to make this change over in Electron.

/cc @paulcbetts

MarshallOfSound:
Fixes #773

### DIFF
--- a/src/Squirrel/UpdateManager.ApplyReleases.cs
+++ b/src/Squirrel/UpdateManager.ApplyReleases.cs
@@ -182,7 +182,7 @@ namespace Squirrel
                         sl.Arguments += String.Format(" -a \"{0}\"", programArguments);
                     }
 
-                    sl.SetAppUserModelId(String.Format("com.squirrel.{0}.{1}", zf.Id, exeName.Replace(".exe", "")));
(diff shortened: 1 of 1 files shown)


paulcbetts:
:+1:, thanks for tracking this down

MarshallOfSound:
@paulcbetts Can we get a patch release with this fix inside it and push it upstream to `electron-winstaller`

Looking to move these fixes through the dependency chains 👍

paulcbetts:
@MarshallOfSound Ya, I'll do it tomorrow

PR #792, opened 2016-08-10
Perform tests against zip path first to help with long paths

damieng:
This tests the zip entry path against the regex to determine whether it is in the lib folder and should be extracted or not.

Previously the code was using the full path which causes paths that were already close to the limit to go over as it also has lib\net45 etc in it.  It also seems that the old code would extract files it shouldn't if the word "lib" appeared in the output path, e.g. `c:\users\libby`)


paulcbetts:
#blessed

PR #803, opened 2016-08-22
Switch to SharpCompress

paulcbetts:
This PR removes old-and-busted ICSharpZipLib and moves us to SharpCompress, which supports 7Zip/LZMA, and is Nice


stefanolson:
Would be really good to be able to use 7Zip, currently using zip my nupkg is 24Mb. With 7Zip it's 15Mb! Will there be an option to use 7Zip to compress?

adamhathcock:
Let me know if there's anything else I can do here to help out as I'm the owner of sharpcompress.

Also, you don't want to use 7Zip.  It's an awful format.  BZip2/LZMA compression is what is desired.

stefanolson:
@adamhathcock I'm just keen to have anything smaller than the current zip format :-)

adamhathcock:
Actually, you might want to consider using LZip from SharpCompress.  @jskeet contributed it

dennismi:
When is this subject to release? any plans on that?

PR #809, opened 2016-08-30
Added optional shutdownAction parameter to UpdateManager.RestartApp

saithis:
For our app Environment.Exit(0) doesn't work, it just crashes your app.
With this optional parameter we can use UpdateManager.RestartApp with our custom shutdown code.

### DIFF
--- a/src/Squirrel/UpdateManager.cs
+++ b/src/Squirrel/UpdateManager.cs
@@ -173,7 +173,7 @@ namespace Squirrel
         }
 
         static bool exiting = false;
-        public static void RestartApp(string exeToStart = null, string arguments = null)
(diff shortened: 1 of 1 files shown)


paulcbetts:
How about an optional boolean parameter `setupOnly` instead that skips Environment.Exit instead

saithis:
Sounds good to me, that way it also works better with async shutdown code.
I'll change it later.

Andrew-Hanlon:
I created a new PR (didn't see this one) with an additional method with the same intention but slightly different semantics. The idea there is to also return the launched Update.exe Process object such that the restart could be 'cancelled' by killing the process.

Issue #737, opened 2016-06-13
TeamCity integration


cmdel:
Perhaps this is documented somewhere but I haven't been able to find it. What's the best way to add the Packaging step in TeamCity? My blocker is the fact that you need to run the squirrely --releasify into the Package Manager Console.
Any thoughts/suggestions on this?

Thanks.

P.S. Apologies, I know this isn't an "issue" but there doesn't seem to be a support forum mentioned anywhere.

DanielRivers:
When you get your code, squirrel will sit under packages if added using nuget.

firstly add a NuGet Pack process which will create the .nupkg, I use a nuspec to ensure the package is right.

Then create a command line build process and add the following:

`%system.teamcity.build.workingDir%\packages\squirrel.windows.1.4.0\tools\squirrel --**releasify [NUPKG PATH]**\%system.build.number%.nupkg -r [OUTPUT PATH]
`

Note: Paths may vary depending on your structure.

cmdel:
Thanks Daniel. Will it not be an issue that this will run outside on the Package Manager Console?

DanielRivers:
No, running this way is the same as running from the package manager console

Thieum:
@shiftkey fixed in #821

PR #821, opened 2016-09-07
Docs Packaging via TeamCity

kenbailey:
Start some basic documentation on automating app packaging via TeamCity (source Issue #737),

Thanks @DanielRivers


dealproc:
Do you gents think maybe a powershell script may be viable here?  I have a starter that can be used as a template to extract a reusable module for build automation.

PR #841, opened 2016-10-02
Allow a controlled exit and app restart

Andrew-Hanlon:
Added a new method that allows the application to handle its own exit (to save settings, documents, etc.) while still making use of the simple app restart semantics. 

The new method returns the Update.exe Process such that the restart could be cancelled (via killing the relaunch process).


paulcbetts:
Thanks!

Issue #843, opened 2016-10-03
Policy issues


LennartAJansson:
We have issues due to execution policy restrictions at our company. We keep a whitelist over folders where execution is ok and would like to be able to define where Update.exe is unpacking and executing. As it is right now it uses %localappdata%\SquirrelTemp with a fallback on C:\ProgramData\%username%\SquirrelTemp, would it be possible to replace SquirrelTemp with some environmentvariable or parameter?

elmahdibenhadda:
Hi,

You can pass the root installation directory to the UpdateManager.
[https://github.com/Squirrel/Squirrel.Windows/blob/master/src/Squirrel/UpdateManager.cs#L34](url)
Hope it helps.

Mehdi.

LennartAJansson:
Tried that:
using (var mgr = new UpdateManager("MyDeployPath",null, "AnotherPath"))
The result was that I got some folders and files in "AnotherPath" but Update.exe was totally blown away so when I try to start the application it only fail saying that it can't find that...

Lennart

LennartAJansson:
Further investigations show:
I have a build and deploy scenario that deploys to a network path, same network path is first parameter to UpdateManager constructor. As a third parameter I have the local folder where we are allowed to execute user-level installations (Clickonce etc).

When I try to perform a clean install on a client protected by the policy it fails with a C++ error related to the Fallback path above (Installation has failed - Failed to extract installer). If I search my folders I can see that it populated %localappdata%\SquirrelTemp with my nupkg's, RELEASES and Update.exe. No trace is found in the allowed path and no log is generated enywhere.

When I try to perform a clean install of the same package on an unprotected server it installs and starts initially BUT not in the path according to the third parameter in the constructor. ALSO, the Update.exe is missing, the shortcut is pointing to the allowed path and there's no Update.exe to be found. The log file is located in the default folder %localappdata%\SquirrelTemp and it shows following in the log:

2016-10-04 09:15:08> ApplyReleasesImpl: About to save shortcut: C:\Users\JohnDoe\Desktop\MyApp.lnk (target C:\Users\JohnDoe\AppData\Local\MyApp\Update.exe, workingDir C:\Users\JohnDoe\AppData\Local\MyApp\app-16.10.4.2, args --processStart "MyApp.exe")

BUT the shortcut that is created contains following Target:
C:\SafeFolder\SquirrelTemp\MyApp\update.exe --processStart "MyApp.exe"

[...]

PR #846, opened 2016-10-05
Added support for environment variable SQUIRREL_TEMP to Setup.exe

LennartAJansson:
Allows to control where Setup.exe will extract and execute the package

#843

### DIFF
--- a/src/Setup/UpdateRunner.cpp
+++ b/src/Setup/UpdateRunner.cpp
@@ -138,7 +138,6 @@ bool CUpdateRunner::DirectoryExistsW(wchar_t* szPath)
 
 bool CUpdateRunner::DirectoryIsWritableW(wchar_t * szPath)
 {
-	try {
 		wchar_t szTempFileName[MAX_PATH];
 		UINT uRetVal = GetTempFileNameW(szPath, L"Squirrel", 0, szTempFileName);
 		if (uRetVal == 0) {
@@ -146,10 +145,6 @@ bool CUpdateRunner::DirectoryIsWritableW(wchar_t * szPath)
 		}
 		DeleteFile(szTempFileName);
 		return true;
-	}
-	catch (...) {
-		return false;
-	}
 }
 
 int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallbackDir)
--- a/src/Setup/UpdateRunner.cpp
+++ b/src/Setup/UpdateRunner.cpp
@@ -168,7 +168,7 @@ int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallback
 			if (DirectoryIsWritableW(envSquirrelTemp)) {
 				if (!PathIsUNCW(envSquirrelTemp)) {
 					_swprintf_c(targetDir, _countof(targetDir), L"%s", envSquirrelTemp);
-					bool envSquirrelTempIsOk = true;
+					envSquirrelTempIsOk = true;
 				}
 			}
 		}
--- a/src/Setup/UpdateRunner.cpp
+++ b/src/Setup/UpdateRunner.cpp
@@ -128,6 +128,30 @@ HRESULT CUpdateRunner::ShellExecuteFromExplorer(LPWSTR pszFile, LPWSTR pszParame
 		CComVariant(SW_SHOWDEFAULT));
 }
 
+bool CUpdateRunner::DirectoryExistsW(wchar_t* szPath)
+{
+	DWORD dwAttrib = GetFileAttributes(szPath);
+
+	return (dwAttrib != INVALID_FILE_ATTRIBUTES &&
+		(dwAttrib & FILE_ATTRIBUTE_DIRECTORY));
+}
+
+bool CUpdateRunner::DirectoryIsWritableW(wchar_t * szPath)
+{
+	try {
+		wchar_t szTempFileName[MAX_PATH];
+		UINT uRetVal = GetTempFileNameW(szPath, L"Squirrel", 0, szTempFileName);
+		if (uRetVal == 0) {
+			return false;
+		}
+		DeleteFile(szTempFileName);
+		return true;
+	}
+	catch (...) {
+		return false;
+	}
+}
+
 int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallbackDir)
 {
 	PROCESS_INFORMATION pi = { 0 };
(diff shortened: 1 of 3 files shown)


LennartAJansson:
Ready to merge @paulcbetts?

src/Setup/UpdateRunner.cpp
```diff
@@ -137,31 +137,37 @@ int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallback
...
-
-		if (!CreateDirectory(targetDir, NULL) && GetLastError() != ERROR_ALREADY_EXISTS) {
-			wchar_t err[4096];
-			_swprintf_c(err, _countof(err), L"Unable to write to %s - IT policies may be restricting access to this folder", targetDir);
-			DisplayErrorMessage(CString(err), NULL);
-
-			return -1;
-		}
+	wchar_t *envSquirrelTemp = _wgetenv(L"SQUIRREL_TEMP");
+	if (envSquirrelTemp) {
```
paulcbetts: We need to check that:

1. This directory exists
2. The user has write access to it
3. It's not a UNC path

LennartAJansson: Ok, I will add that too :)

LennartAJansson: Ok, I will add that too :) Think we should create folder if doesn't exist?

LennartAJansson:
Ok, I will add that too :) Think we should create folder if it doesn't exist? Or just skip using it?

LennartAJansson:
@paulcbetts: Added the stuff you asked for

src/Setup/UpdateRunner.cpp
```diff
@@ -128,6 +128,30 @@ HRESULT CUpdateRunner::ShellExecuteFromExplorer(LPWSTR pszFile, LPWSTR pszParame
...
+	try {
+		wchar_t szTempFileName[MAX_PATH];
+		UINT uRetVal = GetTempFileNameW(szPath, L"Squirrel", 0, szTempFileName);
+		if (uRetVal == 0) {
+			return false;
+		}
+		DeleteFile(szTempFileName);
+		return true;
+	}
+	catch (...) {
```
paulcbetts: What part of this operation could throw a C++ exception?

src/Setup/UpdateRunner.cpp
```diff
@@ -137,31 +161,46 @@ int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallback
...
+					bool envSquirrelTempIsOk = false;
+				}
+			}
+		}
+	}
 
-		SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, SHGFP_TYPE_CURRENT, appDataDir);
-		GetUserName(username, &unameSize);
-		DWORD lastError = GetLastError();
+	if(!envSquirrelTempIsOk) {
```
paulcbetts: Where do we actually use `envSquirrelTemp`?

src/Setup/UpdateRunner.cpp
LennartAJansson: GetTempFileNameW...

src/Setup/UpdateRunner.cpp
LennartAJansson: _swprintf_c(targetDir, _countof(targetDir), L"%s", envSquirrelTemp);
Line 170
And if it's not set we will run the old way

LennartAJansson:
I found a copy and paste bug so envSquirrelTempIsOk was never set to true, sorry for that... Fixed it and pushed

src/Setup/UpdateRunner.cpp
paulcbetts: I don't think that throws C++ exceptions

riverar: This is silly. Permissions could change after this is called, meaning you still have to deal with the failure in the action that proceeds this. Just do the action and skip this check.

src/Setup/UpdateRunner.cpp
```diff
@@ -137,31 +161,46 @@ int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallback
...
+		} else {
+			wchar_t username[512];
+			wchar_t uid[128];
+			wchar_t appDataDir[MAX_PATH];
+			ULONG unameSize = _countof(username);
 
-		_swprintf_c(targetDir, _countof(targetDir), L"%s\\%s", appDataDir, username);
+			SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, SHGFP_TYPE_CURRENT, appDataDir);
+			GetUserName(username, &unameSize);
+			DWORD lastError = GetLastError();
```
riverar: Nit: lastError doesn't seem to be used here.

LennartAJansson: @riverar: That part isn't mine either. I only focused on reading envvariable SQUIRREL_TEMP, it's supported in the managed part of squirrel but was missing in Setup...

src/Setup/UpdateRunner.cpp
LennartAJansson: Calling my attempt to fix the issue with the missing environmentvariable for being silly isn't really constructive @riverar ! I'm trying to contribute since I first asked for a change and if this change isn't going to be made in due time we will simply have to throw squirrel down the drains since we simply can't use it...

src/Setup/UpdateRunner.cpp
```diff
@@ -137,31 +161,46 @@ int CUpdateRunner::ExtractUpdaterAndRun(wchar_t* lpCommandLine, bool useFallback
...
+		if (!useFallbackDir) {
+			SHGetFolderPath(NULL, CSIDL_LOCAL_APPDATA, NULL, SHGFP_TYPE_CURRENT, targetDir);
+		} else {
+			wchar_t username[512];
+			wchar_t uid[128];
+			wchar_t appDataDir[MAX_PATH];
+			ULONG unameSize = _countof(username);
 
-		_swprintf_c(targetDir, _countof(targetDir), L"%s\\%s", appDataDir, username);
+			SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, SHGFP_TYPE_CURRENT, appDataDir);
```
LennartAJansson: Didn't look at that part @riverar, that's old stuff :)

src/Setup/UpdateRunner.cpp
LennartAJansson: That's cool @riverar :) I haven't used C++ since the 90's and never contributed to anything on github either, working with ALM instead of coding nowadays

LennartAJansson:
Ok, I have removed the try - catch part, @paulcbetts

paulcbetts:
Cool, sorry for being picky, we'll get this in today

LennartAJansson:
Awesome @paulcbetts Looks like we could do some more cleaning up in the future according to the comments from @riverar

paulcbetts:
Fixed!

PR #855, opened 2016-10-20
SquirrelAwareApp now supports SemanticVersion

jochenvangasse:
Squirrel Aware apps didn't have support for semantic versioning

If you would pass a semver into a squirrel-aware app, it would just silently fail in the HandleEvents method when trying to make a Version from the semver string.


orthros:
I'm a big fan of Semantic Versioning, but doesn't this break all existing users of Squirrel? I understand you can get a `SemanticVersion` 's `Version`, but this is a relatively major API change.

If it is necessary for you to handle a semantic version in your downstream code, couldn't you do something like the following in your own code:
```csharp
public static void OnInitialInstall(Version version)
{
   OnInitialInstall(new SemanticVersion(version));
}
private static void OnInitialInstall(SemanticVersion version)
{
    ///Do your logic here
}
```

jochenvangasse:
@orthros Hmmm, that might work, but I think we still would have a problem in the HandleEvents method where the second argument from the commandline is fed into a Version constructor

ie.
var version = new Version(args[1]);

orthros:
@jochenvangasse true, but isnt args[1] a string? 
Looking at it deeper, as long as the SemVer you are sending in as a string is compatible with the Version constructor (found [here](https://msdn.microsoft.com/en-us/library/y0hf9t2e(v=vs.110).aspx)), you should be able to construct a compatible SemVer and Version.

That, of course, means that you are limited to the feature set of Version and not the NuGet SemVer ([here](https://github.com/NuGet/Home/wiki/SemVer-2.0.0-support))

So with that in mind I came up with a different solution using some generics and Funcs to pass it down the line. Note, I am **not** that pleased with this overall for two reasons
1. I feel it might cause confusion down the line for users, 
2. We are introducing the NuGet package reference, which I can see has rather large implications down the line in the Update, Update-Mono and the SyncReleases projects

Regardless of these concerns, here is what I've come up with
```csharp

public static void HandleEvents(
    Action<Version> onInitialInstall = null,
    Action<Version> onAppUpdate = null,
    Action<Version> onAppObsoleted = null,
    Action<Version> onAppUninstall = null,
    Action onFirstRun = null,
    string[] arguments = null)
{            
    var lookup = BuildLookup(onInitialInstall, onAppUpdate, onAppObsoleted, onAppUninstall);

    Func<string, Version> selector = new Func<string, Version>((arg) =>
    {
        return new Version(arg);
    });

    HandleEventsImpl(onFirstRun, lookup, selector, arguments);
}

public static void HandleEvents(
    Action<SemanticVersion> onInitialInstall = null,
    Action<SemanticVersion> onAppUpdate = null,
    Action<SemanticVersion> onAppObsoleted = null,
    Action<SemanticVersion> onAppUninstall = null,
    Action onFirstRun = null,
    string[] arguments = null)
{
    Action<SemanticVersion> defaultBlock = (v => { });
            
    var lookup = BuildLookup(onInitialInstall, onAppUpdate, onAppObsoleted, onAppUninstall);

    Func<string, SemanticVersion> selector = new Func<string, SemanticVersion>((arg) =>
        {
            return new SemanticVersion(arg);
        });

    HandleEventsImpl(onFirstRun, lookup, selector, arguments);
}

private static Dictionary<string, Action<T>> BuildLookup<T>(
    Action<T> onInitialInstall = null,
    Action<T> onAppUpdate = null,
    Action<T> onAppObsoleted = null,
    Action<T> onAppUninstall = null) where T : class
{
    Action<T> defaultBlock = (v => { });
    var lookup = new[] {
        new { Key = "--squirrel-install", Value = onInitialInstall ?? defaultBlock },
        new { Key = "--squirrel-updated", Value = onAppUpdate ?? defaultBlock },
        new { Key = "--squirrel-obsolete", Value = onAppObsoleted ?? defaultBlock },
        new { Key = "--squirrel-uninstall", Value = onAppUninstall ?? defaultBlock },
    }.ToDictionary(k => k.Key, v => v.Value);

    return lookup;
}

private static void HandleEventsImpl<T>(Action onFirstRun, 
                                        Dictionary<string,Action<T>> lookup, 
                                        Func<string,T> selector,
                                        string[] arguments = null ) where T : class
{
    var args = arguments ?? Environment.GetCommandLineArgs().Skip(1).ToArray();
    if (args.Length == 0) return;
            
    if (args[0] == "--squirrel-firstrun")
    {
        (onFirstRun ?? (() => { }))();
        return;
    }

    if (args.Length != 2) return;

    if (!lookup.ContainsKey(args[0])) return;
    T version = selector(args[1]);

    try
    {
        lookup[args[0]](version);
        if (!ModeDetector.InUnitTestRunner()) Environment.Exit(0);
    }
    catch (Exception ex)
    {
        LogHost.Default.ErrorException("Failed to handle Squirrel events", ex);
        if (!ModeDetector.InUnitTestRunner()) Environment.Exit(-1);
    }
}

```

orthros:
Another problem I have with my solution is that `Version` and `SemanticVersion` are both `sealed` classes and do not inherit from a common base class, making it difficult, if not impossible to write a satisfactory Generic with them (though I freely admit I am not a Generics wizard like some).

orthros:
Hey @jochenvangasse, I'm thinking about forking this on my own and seeing if I can get more traction on the package building with my proposed changes. If I get anywhere and submit a PR I'll reference this one if you are okay with it?

jochenvangasse:
@orthros Sure, that would be fine. 
I haven't had the time to work on the problem a bit more, but keep me posted.

gojanpaolo:
@shiftkey I believe we can consider closing this PR in favor of PR #865

shiftkey:
Closing in favour of #865

PR #865, opened 2016-10-31
Semantic Versioning Support in SquirrelAwareApp's HandleEvents

orthros:
Inspired by #855 added another overload to SquirrelAwareApp's HandleEvents that takes ```Action<SemanticVersion>``` as opposed to ```Action<Version>``` as arguments.

Rather than re-invent the wheel, the actual implementation of HandleEvents (parse arguments, invoke actions, error reporting) have been abstracted out to the new HandleEventsImpl<T> function, that the two new HandleEvents code calls.

This should maintain backwards compatibility, and allow future and current users of Squirrel.Windows to use ```SemanticVersion```s in their applications

### DIFF
--- a/src/Squirrel/SquirrelAwareApp.cs
+++ b/src/Squirrel/SquirrelAwareApp.cs
@@ -99,7 +99,7 @@ namespace Squirrel
                 return new SemanticVersion(arg);
             });
 
-            HandleEventsImpl(onFirstRun, lookup, selector, arguments);
+            HandleEventsImpl(onFirstRun, lookup, selector, args);
         }
 
         /// <summary>
--- a/src/Squirrel/SquirrelAwareApp.cs
+++ b/src/Squirrel/SquirrelAwareApp.cs
@@ -131,11 +131,11 @@ namespace Squirrel
         {
             Action<T> defaultBlock = (v => { });
             var lookup = new[] {
-                                    new { Key = "--squirrel-install", Value = onInitialInstall ?? defaultBlock },
-                                    new { Key = "--squirrel-updated", Value = onAppUpdate ?? defaultBlock },
-                                    new { Key = "--squirrel-obsolete", Value = onAppObsoleted ?? defaultBlock },
-                                    new { Key = "--squirrel-uninstall", Value = onAppUninstall ?? defaultBlock },
-                                }.ToDictionary(k => k.Key, v => v.Value);
+                new { Key = "--squirrel-install", Value = onInitialInstall ?? defaultBlock },
+                new { Key = "--squirrel-updated", Value = onAppUpdate ?? defaultBlock },
+                new { Key = "--squirrel-obsolete", Value = onAppObsoleted ?? defaultBlock },
+                new { Key = "--squirrel-uninstall", Value = onAppUninstall ?? defaultBlock },
+            }.ToDictionary(k => k.Key, v => v.Value);
 
             return lookup;
         }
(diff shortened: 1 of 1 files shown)


paulcbetts:
This is a good idea but unfortunately it has a few problems:

1. How does the compiler choose which overload to use? The only difference appears to be the <T> of Action<T>
2. It exposes types from our internal NuGet DLL which is probably Not So Good

orthros:
Thanks for taking a look at this and pointing those out!

1. Truth be told, I'm not positive what your question here is referring to. From the high level, the compiler knows which method to call because the two overloaded methods are strongly typed. Consider the following code:
```csharp
using NuGet;
using Squirrel;
using System;

namespace TestCAppSquirrel
{
    class Program
    {
        static void Main(string[] args)
        {
            SquirrelAwareApp.HandleEvents(null, SemVOnAppUpdate, null, null, null, null);

            SquirrelAwareApp.HandleEvents(null, VOnAppUpdate, null, null, null, null);

            Console.ReadKey();
        }        

        public static void SemVOnAppUpdate(SemanticVersion version)
        {
            Console.WriteLine("The SemanticVersion is " + version);
        }

        public static void VOnAppUpdate(Version version)
        {
            Console.WriteLine("The Version is " + version);
        }        
    }
}
```
This compiles with the `HandleEvents` using `SemanticVersion` and passes the `Action<SemanticVersion>` through correctly
image
Uses the selector to create a `SemanticVersion`
image
And makes it to the proper `Action`
image

Similarly for the `Version` overload of this code.

2. I agree on this point, but when I attempted to use standard NuGet `SemanticVersion` distributed via the  Package Manager, I had no end of compiling errors (as you might imagine). I'll try again shortly after writing this. 

Reporting back about 20 minutes later. I managed to get this to compile and run by adding a reference to `NuGet.Core` in the Squirrel project. This conflicts with the internal `NuGet` package so I used an `extern alias` to get it to work (learning something new every day 👍 )
```csharp
extern alias NuGetCore;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using Splat;
using NuGetCore::NuGet;
```
This is not the most elegant solution, but it doesn't expose the internal `NuGet` type to users. The problem we run into when doing this is now every user of Squirrel that wants to use a `SemanticVersion` will need to first reference the `NuGet.Core` package from NuGet, then they too will need to ExternAlias the `NuGet.Core` package in order to compile (which I had to do in my test console app). From a user-experience standpoint, this does not seem optimal, but that's the trade-off we get by not wanting to expose the internal libraries. 

On that note, are we not already exposing this to our users already? I made another test console app from scratch, added the Squirrel.Windows package, and could immediately use the `SemanticVersion` from the internal `NuGet` package, so I'm not positive this is "exposing" more than we already are.
image


Looking forward to hearing your thoughts!

orthros:
As a note: I took the code I wrote to reference NuGet proper and committed it to another branch [hide-internal-nuget](https://github.com/orthros/Squirrel.Windows/tree/hide-internal-nuget). With the principal commit in question here  17652334eb2f203bb925ad8c56e4ef16b43a89aa

orthros:
So I was thinking about this over the weekend and I came to this conclusion:

We are already exposing our internal implementation of the NuGet Package to users when we deploy Squirrel via NuGet, so any Squirrel Users that are also using the official NuGet package are already using the `extern alias` feature to differentiate the DLLs.

By us referencing the official NuGet package, we will then force *all* existing users of Squirrel to use the `extern alias` feature.

For the sake of backwards compatibility and Ease of Use for existing users, I say we should keep using the Internal `NuGet` library and provide some strong documentation on the less likely case of a DLL conflict.

vadi2:
@anaisbetts Is it possible to merge this in?

PR #884, opened 2016-11-25
Fix for detecting squirrel aware app when Mono.Cecil does not detect Custom Attributes

inderpal58:
Installation and Updation was working fine until exe was not executed through ConfuserEx.

When exe was processed by ConfuserEx; shortcuts was created for all exe in folder.
The reason was Mono.Cecil says no custom attributes.
When used system.reflection; custom attributes was detected successfully.

jbevain/cecil#305

### DIFF
--- a/src/Squirrel/SquirrelAwareExecutableDetector.cs
+++ b/src/Squirrel/SquirrelAwareExecutableDetector.cs
@@ -38,8 +38,8 @@ namespace Squirrel
             try {
                 var assembly = AssemblyDefinition.ReadAssembly(executable);
                 if (!assembly.HasCustomAttributes) {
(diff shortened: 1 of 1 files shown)


paulcbetts:
Hey @inderpal58, sorry I'm taking forever to respond. So, the reason that we use Mono.Cecil is that if you call `Assembly.LoadFile`, you will load that assembly into Squirrel, which means that it has to be _able_ to load (i.e. all of its dependencies have to work in .NET 4.5, etc etc etc). This is how the original Squirrel.Windows worked and in practice it was super faulty.

Maybe you can figure out how to use Mono.Cecil to get the attributes you need?

PR #886, opened 2016-11-28
modify the providedClient to a webClientFactory in FileDownloader

### LINKED ISSUE #880: FileDownloader's providedClient is used concurrently thus can throw "WebClient does not support concurrent I/O operations"
I'm getting exceptions originating from `UpdateManager.DownloadReleases` with the following message

    WebClient does not support concurrent I/O operations

and it's because I use my own ✨WebClient✨ instance.

In the `FileDownloader`'s constructor a `WebClient providedClient` can be specified. Because [UpdateManager.DownloadReleases](https://github.com/Squirrel/Squirrel.Windows/blob/master/src/Squirrel/UpdateManager.DownloadReleases.cs#L33) downloads the packages parallel, `FileDownloader.DownloadFile` can be called concurrently, which uses a "singleton" WebClient concurrently, however WebClient only supports a single operation, it cannot download multiple files at once, that's why it's throwing exceptions.

The file downloader works by default because `Utility.CreateWebClient();` is called on every `DownloadFile`.

    public class FileDownloader : IFileDownloader, IEnableLogger
    {
        private readonly WebClient _providedClient;

        public FileDownloader(WebClient providedClient = null)
        {
            this._providedClient = providedClient;
        }

        public async Task DownloadFile(string url, string targetFile, Action<int> progress)
        {
            WebClient wc = this._providedClient ?? Utility.CreateWebClient();
            
            ...

I can workaround this in my codebase, but I'd like to change it in Squirrel to something like this below. What do you think?

    public class FileDownloader : IFileDownloader, IEnableLogger
    {
        private readonly Func<WebClient> _webClientFactory;

        public FileDownloader(Func<WebClient> providedClientFactory = null)
        {
            this._webClientFactory = providedClientFactory ?? Utility.CreateWebClient
        }

        public async Task DownloadFile(string url, string targetFile, Action<int> progress)
        {
            WebClient wc = _webClientFactory(); 
            
            ...

szanto90balazs:
fixes #880

so for every file download there's a newly created WebClient instance, so FileDownloader can be used concurrently

### DIFF
--- a/src/Squirrel/FileDownloader.cs
+++ b/src/Squirrel/FileDownloader.cs
@@ -13,16 +13,16 @@ namespace Squirrel
 
     public class FileDownloader : IFileDownloader, IEnableLogger
     {
-        private readonly WebClient _providedClient;
+        private readonly Func<WebClient> _webClientFactory;
 
-        public FileDownloader(WebClient providedClient = null)
+        public FileDownloader(Func<WebClient> webClientFactory = null)
         {
-            _providedClient = providedClient;
+            _webClientFactory = webClientFactory ?? Utility.CreateWebClient;
         }
 
         public async Task DownloadFile(string url, string targetFile, Action<int> progress)
         {
-            using (var wc = _providedClient ?? Utility.CreateWebClient()) {
+            using (var wc = _webClientFactory()) {
                 var failedUrl = default(string);
 
                 var lastSignalled = DateTime.MinValue;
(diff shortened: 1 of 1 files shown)


paulcbetts:
This changes a public signature so it's a breaking change, can you support the old behavior as well?

szanto90balazs:
Sure, I also made it a bit more bulletproof.

phw198:
Will this change be getting merged? I'm also hitting #880 (Squirrel v1.9.0).

Issue #77, opened 2014-09-30
Customize the time before the loading GIF shows up


christianrondeau:
It's currently hardcoded to 4 seconds, which I think is way too much. I expect immediate feedback, and even though the app might start immediately, just the .NET JIT may be enough to make me ask myself the question: what is happening? Did it work?

I would personally set it to zero, and even if it just flickers, at least it gives a clue that something happened.

kevfromireland:
I would be up for making this configurable. I think I understand the rationale between not wanting it until after 4s by default, but would you accept a PR that made it opt-in configurable with 4s as the default?

paulcbetts:
> I would be up for making this configurable. I think I understand the rationale between not wanting it until after 4s by default, but would you accept a PR that made it opt-in configurable with 4s as the default?

I'll probably lower this to maybe 2sec, but I don't really want it to be configurable

daviseford:
+1 to this issue, I'd really like to see the ability to customize when the loadingGif is shown! Not being able to show it early makes users freak out about what's happening.

demetris-manikas:
@paulcbetts Is there a reason that you don't want a hardcoded value to be configurable apart from having to write code to handle it?

paulcbetts:
@demetris-manikas I'd rather just have code that works for everyone without configuration, especially because we also have no place to currently put this configuration

develar:
I am with @paulcbetts but unfortunately windows is windows and we should show immediate response to user action — on OS X user after click get "Verifying" (DMG) or bouncing app icon in the dock (.app). So... when I will have time I will fix this issue to close https://github.com/electron-userland/electron-builder/issues/374/

demetris-manikas:
`especially because we also have no place to currently put this configuration`
Fair enough.
@paulcbetts Still I don't get why an installer should not show anything to the user if the installation is fast. I am on the 0 seconds delay side and since users complain the code does not  just work for everybody.
Thanks anyway.

develar:
@demetris-manikas 0 time delay break "just run it" behaviour. If on click such progress indicator will be shown immediately and then quickly disappears — it is not good. Well, since it is windows, it may be ok, but ideally should be no progress indicator if installation performed quickly.

> I'll probably lower this to maybe 2sec

I think, we can try to do it. In any case for me current behaviour is ok — well, I run windows in a virtual machine, but even in this case no noticeable delay.

demetris-manikas:
[...]

Issue #878, opened 2016-11-21
User confusion about Setup


ZARk-be:
I recently posted a bug about "some" users not being able to update.

Actually what happened, is that even though the file is called setup.exe, those users thought the app was standalone.

Quoting : 
"Not being an installer, I just popped the file on my desktop"
"Done that, although that setup file is just the executable."

Users expect a setup file to at least warn them that the app was installed and a shortcut created.
I will force a splash image to show. 
As right now, 50% of the beta testers had that issue.
(It doesn't help that the very first version of my app was portable i guess).

Don't get me wrong, i love the idea of a fast install (nothing worse than a setup app that don't start the exe after setup ... *sigh*), but a quick splash image would avoid all possible confusion i think.

dealproc:
Can you use the splash installation features of Squirrel to achieve it?  (e.g. they do it for Atom.IO text editor)

ZARk-be:
> Squirrel will optionally display a graphic as a "splash screen" while installation is processing, but only if installation takes more than a pre-set amount of time

we should be able to force the splash screen

hvassbotn:
Agree about forcing the splash screen. We have a largish install (45 MB), and I know that it will take time to install and eventually show the progress splash, so it would be nice to force it from the beginning (while the user wonders if nothing is happening)...

ZARk-be:
I've forked it in my github, and have added an option to specify delay. It flashes on mine, so you can barely read it. It's 4seconds by default, so i kept that.

Thinking of adding another option to show another gif after successfull installation, where the user must click the image to close it, and continue with launching the app.

The ONLY way i see of telling the use the program was successully installed, is to make the app SquirrelAware and show a special dialog box in the app after successfull install.... 

I'll create a pull request afterwards, we'll see if they appreciate the idea, even though it's kind of against the whole point of Squirrel, to be invisible and non-intrusive( which is a good thing), but users aren't ready yet for this, they still need a transition from the god-awfull wizard installations to the invisible squirrel.

gojanpaolo:
@ZARk-be If I understand correctly, the original issue (user confusion about Setup) is solved by including a splash screen. Can we close this issue?

PR #889, opened 2016-12-02
-d option can now specify the delay before showing splash screen.

ZARk-be:
As suggested here https://github.com/Squirrel/Squirrel.Windows/issues/878 , the delay before showing the SplashScreen shouldn't be hardcoded, so i put it as an option in Milliseconds. It defaults to the previous hard-coded value, so it's not a breaking change.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -87,6 +87,7 @@ namespace Squirrel.Update
                 string packagesDir = default(string);
                 string bootstrapperExe = default(string);
                 string backgroundGif = default(string);
+                int backgroundGifDelay = 4000;
                 string signingParameters = default(string);
                 string baseUrl = default(string);
                 string processStart = default(string);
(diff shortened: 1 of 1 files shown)


paulcbetts:
Who's the audience for this feature? Like, who is running setup.exe via the command line and providing this parameter?

ZARk-be:
This is a command line parameter for squirrel --releasify .
https://github.com/Squirrel/Squirrel.Windows/blob/master/docs/using/squirrel-command-line.md

you can do a -g .\loading.gif and now add -d 1000  to have the splashscreen show after 1second instead of the default 4 sec.
Or -d 0  if you want it to display asap.

paulcbetts:
@ZARk-be ...that's not what this PR does though

vadi2:
I'd like to see this in, setup being too quick affects our application as well. @ZARk-be any updates?

gojanpaolo:
This will fix #77 when merged

fpadilha:
This should be in... We've done some testing (actually watching people create an account, download, install and use our software) and the results about that 4-second delay are horrible. Almost everyone tries to open the setup twice.

PR #924, opened 2017-01-12
Fixed semenatic versioning regex to support uppercase characters.

runehalfdan:
From SemVer 1.0: http://semver.org/spec/v1.0.0.html
The string MUST be comprised of only alphanumerics plus dash [0-9A-Za-z-]. Pre-release versions satisfy but have a lower precedence than the associated normal version.


paulcbetts:
:+1: thanks for the fix

PR #928, opened 2017-01-17
Support semantic versioning in SquirrelAwareApp.HandleEvents using Action<string>

runehalfdan:
Added overloads to SquirrelAwareApp.HandleEvents that takes string instead of Versions.

This is because .NET Version-class don't support semantic version strings, ie. 2.0.1-patch123

I considered marking the HandleEvents taking Action<Version> as obsolete since SemanticVersion is use throughout the rest of Squirrel, but did not want to risk introducing warnings


runehalfdan:
I think this way of supporting semantic versions is better that adding overloads that take SemanticVersion to avoid adding dependencies on other packages. 

Also (for us) the version-argument is only used for logging so we don't have any benefit from taking Version/SemanticVersion in those callbacks.

gojanpaolo:
@shiftkey This is another approach to add support for semantic versioning (see PR #865 for supporting `Nuget.SemanticVersion`). Might be a good idea to change the title to something like "Support semantic versioning in `SquirrelAwareApp.HandleEvents` using `Action<string>`".

PR #932, opened 2017-01-23
Fix for missing registry key

Scyssion:
Fixes crash that may occur if the user does not have the expected registry key.


Scyssion:
Just something I came across while debugging.  A simple fix.

Issue #919, opened 2017-01-07
Version number issue when packaging the "MsBuild target" way


Mario-Eis:
The issue is related to the procedure described here: https://github.com/Squirrel/Squirrel.Windows/blob/master/docs/using/visual-studio-packaging.md

When GetAssemblyIdentity gives a version like 1.0.1.0, nuget internally seems to use 1.0.1 to create the nupkg file. So the created files name is "MyApp.1.0.1.nupkg". The call to squirrel uses 1.0.1.0 in the filename to look it up, because thats what GetAssemblyIdentity returns.
Only way to get around this is to ALWAYS have at least a 1 for the revision number. With 1.0.1.1 nuget creates a "MyApp.1.0.1.1.nupkg"

Thieum:
@shiftkey there's a PR open for this, further discussion could take place there, and this can be closed.

PR #942, opened 2017-02-02
Version number issue when packaging the "MsBuild target" way

ChrisPulman:
Suggested fix for #919
Updated Visual-Studio-packaging to show a Semantic method that should be compatible with the current version of Squirrel


pltaylor:
I was about to do a pull request with a simpler solution of just replacing the current lines w/ $([System.Version]::Parse(%(myAssemblyInfo.Version)).ToString(3)) instead of %(myAssemblyInfo.Version), it appears to do the same thing with a lot less work.

ChrisPulman:
@pltaylor thanks, I have updated the `ITask` to use `ToString(3)`

PR #948, opened 2017-02-13
Ensure a console has been allocated before using it

BillHally:
When an exception is raised by Program.main, Program.Main will attempt to write it out to the console before returning with the exit code of -1.

However, if the console has not yet been allocated, then the user cannot see the error. This pull request ensures that the console has been allocated before the exception is written out to the console.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -31,9 +31,10 @@ namespace Squirrel.Update
             try {
                 return pg.main(args);
             } catch (Exception ex) {
+                ensureConsole();
+                Console.Error.WriteLine(ex);
                 // NB: Normally this is a terrible idea but we want to make
                 // sure Setup.exe above us gets the nonzero error code
-                Console.Error.WriteLine(ex);
                 return -1;
             }
         }
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -31,10 +31,9 @@ namespace Squirrel.Update
             try {
                 return pg.main(args);
             } catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


paulcbetts:
This means that when things fail, random DOS boxes pop up, that seems Bad. We only want to call ensureConsole in places where we know the command was invoked by a developer, not as part of the installation

BillHally:
That's a good point.

I've moved the call to `ensureConsole` to the start of `Releasify`.

paulcbetts:
:+1:

lukeskinner:
In my testing of this I can't get any console messages to appear. The following code is present in the ensureConsole() method:
```
NativeMethods.GetStdHandle(StandardHandles.STD_ERROR_HANDLE);
NativeMethods.GetStdHandle(StandardHandles.STD_OUTPUT_HANDLE);
```

The GetStdHandle returns a IntPtr which is never used.

I got it working by adding the following for both handles - I don't know if this is the desired way of handling this so I haven't submitted a PR for it yet:

```
            IntPtr stdHandle = NativeMethods.GetStdHandle(StandardHandles.STD_OUTPUT_HANDLE);
            SafeFileHandle safeFileHandle = new SafeFileHandle(stdHandle, true);
            FileStream fileStream = new FileStream(safeFileHandle, FileAccess.Write);
            Encoding encoding = Encoding.Default;
            StreamWriter standardOutput = new StreamWriter(fileStream, encoding);
            standardOutput.AutoFlush = true;
            Console.SetOut(standardOutput);
```

PR #949, opened 2017-02-13
fix issue where SyncReleases wouldn't work with TLS1.2

TakeN0:
This is the same issue that was fixed with https://github.com/Squirrel/Squirrel.Windows/commit/b464b41bb243380c5acad82f3336ff7c9b1e4cb3 but since SyncReleases does not use the WebClient object the fix did not apply.


paulcbetts:
:+1:

Issue #798, opened 2016-08-12
Firewall access request when app get updated.


inderpal58:
Whenever fresh client is installed the request is generated for firewall access.  On new install its acceptable.

As the application is updated, the exe location is changed and again the user get a prompt for firewall access. That seems irritating.

Is there any solution for this problem?

arul-gupta:
@paulcbetts Is #868 going to help with this issue of firewall prompts on each update?

profix898:
Having the same issue here (with firewall complaining after each update). #951 seems to be on the right track for solving this.

peters:
#1052

PR #951, opened 2017-02-16
Run from same folder

arul-gupta:
This PR tries to solve #798 by always running the app from the same folder(named current).

After every update, the packages are extracted to a temp folder along with the new versioned folder. When update.exe restarts the application, it checks for this temp folder. If it exists, it copies to a folder named current and runs the application from current.
If for some reason, it cannot find current, it falls back to the finding the latest versioned folder.
The stub executable does the same and first checks for current before trying to find the latest folder.

### DIFF
--- a/src/StubExecutable/StubExecutable.cpp
+++ b/src/StubExecutable/StubExecutable.cpp
@@ -44,6 +44,17 @@ std::wstring FindLatestAppDir()
 {
 	std::wstring ourDir;
 	ourDir.assign(FindRootAppDir());
+	
+	//If current exists, just use that
+	std::wstring currDir;
+	currDir.assign(FindRootAppDir());
+	currDir += L"\\current";
+
+	WIN32_FIND_DATA currInfo = { 0 };
+	HANDLE currFile = FindFirstFile(currDir.c_str(), &currInfo);
+	if (currFile != INVALID_HANDLE_VALUE) {
+		return currDir.c_str();
+	}
 
 	ourDir += L"\\app-*";
 
--- a/src/Squirrel/UpdateManager.ApplyReleases.cs
+++ b/src/Squirrel/UpdateManager.ApplyReleases.cs
@@ -285,6 +285,7 @@ namespace Squirrel
             {
                 return Task.Run(async () => {
                     var target = getDirectoryForRelease(release.Version);
+                    var currentTemp = new DirectoryInfo(Path.Combine(rootAppDirectory, "currentTemp"));
 
                     // NB: This might happen if we got killed partially through applying the release
                     if (target.Exists) {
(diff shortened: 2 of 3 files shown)


paulcbetts:
This PR is conceptually flawed, you can't have a folder named "current" then try to copy into it while the app is still running, that's the whole reason we have separate folders, because you can't replace binaries that are in-memory.

arul-gupta:
> because you can't replace binaries that are in-memory.

That's why the package is initially extracted in a temp folder and when Update.exe tries to restart the app, I wait for the parent to exit before moving that temp folder to current.

I've already integrated this in my application, as firewall prompts are just not acceptable after every update. My app runs in sort of kiosk mode and is controlled remotely with the assumption that there is no mouse connected to the system after the initial install.

inderpal58:
One solution may be to add firewall entry manually into registry before app starts. Need to verify....

peters:
#1052

src/StubExecutable/StubExecutable.cpp
```diff
@@ -93,20 +106,32 @@ int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
...
 	si.wShowWindow = nCmdShow;
 
-	std::wstring cmdLine(L"\"");
+	/*std::wstring cmdLine(L"\"");
 	cmdLine += fullPath;
 	cmdLine += L"\" ";
 	cmdLine += lpCmdLine;
+	*/
+
+	std::wstring cmdLine(L"\"");
```
bddckr: When passing **multiple** arguments to your Squirrel-aware up that then is restarting to update this block of code needs to change a little:
```c++
	std::wstring cmdLine(L"\"");
	cmdLine += updateFile;
	cmdLine += L"\" --processStart ";
	cmdLine += appName;
	cmdLine += L" -a=\"";
	cmdLine += lpCmdLine;
	cmdLine += L"\"";
```
We're basically wrapping the arguments to pass with `"` :)

sandeep1995:
@arul-gupta We are using this PR for a while. 
For some users, we are having the following issue. Do you have any solution?
```
2019-12-03 19:11:59> Program: Failed to empty current directory, will try to override files
2019-12-03 19:11:59> Program: Failed to move current directory: System.IO.IOException: The process cannot access the file 'C:\Users\User\AppData\Local\XXXXX\current\XXXXX.exe' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalCopy(String sourceFileName, String destFileName, Boolean overwrite, Boolean checkHost)
   at System.IO.FileInfo.CopyTo(String destFileName, Boolean overwrite)
   at Squirrel.Utility.CopyDirectory(DirectoryInfo source, DirectoryInfo target)
   at Squirrel.Update.Program.RunFromCurrent(String exeName, String appDir, FileInfo targetExe)
```

tkouba:
> This PR is conceptually flawed, you can't have a folder named "current" then try to copy into it while the app is still running, that's the whole reason we have separate folders, because you can't replace binaries that are in-memory.

I'm considering of using junction (soft link) named current to a "latest" folder, see https://docs.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions#junctions. Would that be the solution?

PR #955, opened 2017-02-19
fix verson parse exception on HandleEvents when SemVer(i.e. 1.0.1-rc)…

CodeFunta:
SquirrelAwareApp.HandleEvents throws exception if used SemVer (i.e. 1.0.1-rc) since Version can't parse it


paulcbetts:
Good fix

jleclanche:
@paulcbetts Is squirrel getting a new release soon with this fix? This seems like a pretty major issue :)

PR #993, opened 2017-04-06
Update faq.md

Trinitek:
Fixed formatting issue with code block


Trinitek:
An existing fix is already in an open pull request. My bad.

PR #1001, opened 2017-04-12
Stop downloading remote updates that are lower than the current version

ilyalopatkin:
If the current version is not present on the update server then currently it will download the latest available (which is older), unpack it, and then after an app restart clean up by deleting it. This change makes sure old updates are not considered thus simply skipping the update process. This is needed for running the newly built versions (for testing purposes) which haven't been uploaded to the update server yet.

### DIFF
--- a/src/Squirrel/UpdateInfo.cs
+++ b/src/Squirrel/UpdateInfo.cs
@@ -63,7 +63,7 @@ namespace Squirrel
                 return new UpdateInfo(null, new[] { latestFull }, packageDirectory);
             }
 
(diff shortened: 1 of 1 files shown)


paulcbetts:
Thanks @ilyalopatkin!

Mr-Wallet:
Is it now impossible to "downgrade" clients as a result of this change? i.e. if there is a catastrophic release that requires an immediate revert, is the only fix to re-build the prior version with a higher semver?

anaisbetts:
Correct

Mr-Wallet:
Thank you for the quick response!

PR #1012, opened 2017-05-02
Fixes formatting for Visual Studio Packaging docs

PHeonix25:
Unfortunately, three tildes works as ~~~strike-through~~~ in GFM, and not as "code-sample".
As a result, this was breaking the formatting and layout of the bottom of the page.
To fix this I have simplified the NuGet Package Manager code sample to use the standard GFM "code" convention.

_NOTE: using the GitHub online editor to make this small change meant that the windows new-lines (^M) got replaced with /r/n - I've left those changes in, but I'm happy to remove it if required._


paulcbetts:
Thanks!

Issue #916, opened 2017-01-05
Invalid Product Version - Override to enable larger product versions


simquad:
Hi. Great project, all that jazz.

From my understanding ProductVersions (as specified here: https://msdn.microsoft.com/en-us/library/windows/desktop/aa370859(v=vs.85).aspx) are only allowed to fall within this range:
`Product version must have a major version less than 256, a minor version less than 256, and a build version less than 65536.`
![image](https://cloud.githubusercontent.com/assets/820625/21678238/dd8c6fd0-d335-11e6-8844-bdfcf36f96c3.png)

Now, I know I can set a ProductVersion to higher than this, and it works.
![image](https://cloud.githubusercontent.com/assets/820625/21678223/cde36ca0-d335-11e6-9c14-f1436916d76a.png)

However, it would seem Wix/Candle does not allow this to be set.

The version number system we are using, includes current year and sprint number; it may not be a system for everyone, but it fits us.

I'm trying to understand if it is possible to overwrite this restriction. I've tried to find documentation to assist this (and then aim to send in an appropriate PR), but to no avail.

Any help would be incredible.

Thanks.
Rich

paulcbetts:
No idea, but version numbers in Squirrel need to be Semver compliant so keep that in mind. This is probably a better question for the WiX people, if you figure it out I'll take a PR

simquad:
Thanks Paul.

I'm going to take a look at what is provided in this post (https://sourceforge.net/p/wix/mailman/message/25481068/) and see if anything obvious comes up.

rcorredera:
Hi, I'm using WIX in a project building my own MSI and as they say in https://wixsharp.codeplex.com/discussions/576142 you can bypass the problem by using 
bind.FileVersion.
Explained here : http://stackoverflow.com/questions/626033/how-can-i-set-the-wix-installer-version-to-the-current-build-version
Can you try to use it in the Setup.wxs generated ?

PR #1023, opened 2017-05-15
Issue 916

rcorredera:
See https://wixsharp.codeplex.com/discussions/576142 and http://stackoverflow.com/questions/626033/how-can-i-set-the-wix-installer-version-to-the-current-build-version.

Refs #916


paulcbetts:
Thanks!

Issue #637, opened 2016-03-23
GetShortcutsForExecutable doesn't set the ShortCutFile path


LaviniaC:
It calculates, logs it but does not set it.

paulcbetts:
ShortCutFile is only set when the shortcut is already existing, and GetShortcutsForExecutable returns the shortcut that _would_ be created. Or are you saying that you want to know where we _would_ put the shortcut?

LaviniaC:
Yes, I would. 
In my particular case I just want to change a small part of it (a rename).

paulcbetts:
@LaviniaC I think you'll probably have to return it as a separate thing, but I'm totally :+1: for the idea. Can you submit a PR to this effect?

DanielRivers:
I am not sure that it is setting set if already existing ...

`mgr.CreateShortcutsForExecutable("Startup.exe", ShortcutLocation.Desktop | ShortcutLocation.StartMenu, false, null, null);`
`var test = mgr.GetShortcutsForExecutable("Startup.exe", ShortcutLocation.Desktop | ShortcutLocation.StartMenu | ShortcutLocation.Startup, null).ToList();`

I run the above lines, and still in the GetShortcutsForExecutable results ShortCutFile is empty

PR #1037, opened 2017-05-27
Added support for dashes in installation path.

### LINKED ISSUE #1035: Dashes in Package Name causing bad installation path
It seems that dashes are causing issues too. I've been testing it and it causes the installation path to be truncated: if your package Id is `myapp-staging` the installation path will be truncated to 
 `%appdata%\Local\myapp` causing issues if you have also `myapp-production` and so on.

See [ReleaseEntry.cs#68](https://github.com/Squirrel/Squirrel.Windows/blob/14d6dc1599676a9d83daba60e73507d12b73164d/src/Squirrel/ReleaseEntry.cs#L68). Here is where the name of the folder package name is created and all dashes and dots are removed. 

Then, [here](https://github.com/Squirrel/Squirrel.Windows/blob/master/src/Update/Program.cs#L202) the `PackageName` property is used to build the installation path.

I guess that dashes are removed because the filename is retrieved from the RELEASES file and thus getting a filename similar to this `myapp-name-1.2.4-full.nupkg`.

My approach would be to change the `PackageName` property implementation and use a regular expression.

I'll be willing to submit a PR if it's ok.

robertohuertasm:
Fixes #1035.

The issue was also raised in [electron-builder`s repo #637](https://github.com/electron-userland/electron-builder/issues/637)

### DIFF
--- a/src/Squirrel/ReleaseEntry.cs
+++ b/src/Squirrel/ReleaseEntry.cs
@@ -63,9 +63,14 @@ namespace Squirrel
         [IgnoreDataMember]
         public SemanticVersion Version { get { return Filename.ToSemanticVersion(); } }
 
+        static readonly Regex packageNameRegex = new Regex(@"^([\w-]+)-\d+\..+\.nupkg$");
         [IgnoreDataMember]
         public string PackageName {
-            get { return Filename.Substring(0, Filename.IndexOfAny(new[] { '-', '.' })); }
+            get
+            {
+                var match = packageNameRegex.Match(Filename);
+                return match.Success ? match.Groups[1].Value : Filename.Substring(0, Filename.IndexOfAny(new[] { '-', '.' }));
+            }
         }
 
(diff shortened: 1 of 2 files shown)


src/Squirrel/ReleaseEntry.cs
```diff
@@ -65,7 +65,12 @@ protected ReleaseEntry(string sha1, string filename, long filesize, bool isDelta
 
         [IgnoreDataMember]
         public string PackageName {
-            get { return Filename.Substring(0, Filename.IndexOfAny(new[] { '-', '.' })); }
+            get
+            {
+                var rx = new Regex(@"^([\w-_\s]+)-\d+\..+\.nupkg$");
```
paulcbetts: Can you pull this out into a constant? No reason to recreate it all the time

robertohuertasm: I've added a private static var with the regex.

paulcbetts:
Thanks!

PR #1051, opened 2017-06-07
Only generate execution stubs for the top-level executables

paulcbetts:
This PR makes it so we don't generate execution stubs for subfolder files, which is Usually Unwanted:tm:

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -406,7 +406,7 @@ namespace Squirrel.Update
 
                             this.Log().Info("About to sign {0}", x.FullName);
                             await signPEFile(x.FullName, signingOpts);
-                        })
+                        }, 1)
                         .Wait();
                 });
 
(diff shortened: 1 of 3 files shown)


src/Update/Program.cs
```diff
@@ -405,7 +406,7 @@ public void Releasify(string package, string targetDir = null, string packagesDi
 
                             this.Log().Info("About to sign {0}", x.FullName);
                             await signPEFile(x.FullName, signingOpts);
-                        })
+                        }, 1)
```
damieng: What's the ,1 for?

paulcbetts: `1 /* at a time */`

damieng: Ahhh.  That's the intermittent concurrency problem! Nice catch.

paulcbetts: Yeah, normally it will do 4 at a time, but apparently SignTool is Not Down with that

Issue #1053, opened 2017-06-08
Add option to processStart in order to hide child process window


QJonny:
Hi, my problem is that when using Update.exe to start a child process, the latter being a console application, the console is shown.
Even if the child process hides its console, the latter is shown for a very brief time. I would like to completely hide the console.

It would be very nice to have an option when running the processStart command in order to hide this console (ex: --no-window).

Thanks!!

paulcbetts:
processStart is actually deprecated, you should use the stub executables instead, which will respect the nShowWindow params passed to it

Skywalker13:
Hi @paulcbetts,

It's fine in the case of .NET apps and electron apps because squirrel is available for C# and JS. But for other apps (c/c++/etc,...) how is possible to have an update without using `Update.exe --processStart` ? The stub seems just looking in the right version directory in order to start the app; `Update.exe` is never used in this case then the app will never been updated.

Or maybe I'm wrong ?!?

paulcbetts:
Squirrel in Electron just shells to `update.exe`. Are you saying you want to use the process start and wait feature still? That actually might _not_ be deprecated and you're right and sorry I forgot about this

PR #1054, opened 2017-06-08
Add option in order to hide window of child process during processSta…

QJonny:
…rt. Close issue #1053


paulcbetts:
Thanks for the PR but see #1053 for the resolution, stub executables handle this correctly

PR #1057, opened 2017-06-09
DeltaCompressionDotNet dependency version mismatch

rschoenbach:
The DeltaCompressionDotNet dependency requirement is lower than the requirement set by Squirrel project file.


paulcbetts:
Thanks!

peppy:
Could you please consider pushing a build with this; the current live nuget packages are very broken.

PR #1061, opened 2017-06-14
Fix memory issues with certain Animated GIFs

paulcbetts:
This PR fixes an issue where certain GIFs would immediately cause all available memory to be allocated, which Seems Bad

/cc @niik @shiftkey


src/Update/Update-Mono.csproj
```diff
@@ -70,16 +70,11 @@
...
-    <Reference Include="WpfAnimatedGif, Version=1.4.14.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\WpfAnimatedGif.1.4.14\lib\net\WpfAnimatedGif.dll</HintPath>
-      <Private>True</Private>
-    </Reference>
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\SolutionAssemblyInfo.cs">
       <Link>Properties\SolutionAssemblyInfo.cs</Link>
     </Compile>
-    <Compile Include="AnimatedGifWindow.cs" />
```
shiftkey: So no Mono support for this? Or did this not work at all before?

src/Update/TaskbarManager.cs
```diff
@@ -0,0 +1,251 @@
...
+
+        /// <summary>
+        /// Indicates whether this feature is supported on the current platform.
+        /// </summary>
+        public static bool IsPlatformSupported
+        {
+            get
+            {
+                // We need Windows 7 onwards ...
+                return true;
```
shiftkey: Worth actually checking here? Or is the "no parents" approach Good Enough?

src/Update/AnimatedGifWindowWinForms.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            pictureBox = new PictureBox();
+            this.Controls.Add(pictureBox);
+
+            if (File.Exists(source)) {
+                pictureBox.ImageLocation = source;
+            }
+
+            this.WindowState = FormWindowState.Minimized;
+            Action size = () => { pictureBox.Width = this.Width; pictureBox.Height = this.Height; pictureBox.Left = 0; pictureBox.Top = 0; };
```
shiftkey: omg (lack of) indenting

src/Update/Update-Mono.csproj
paulcbetts: Mono is only for building packages (aka `--releasify`), no changes here

src/Update/TaskbarManager.cs
paulcbetts: .NET 4.5. is >= Win7 only anyways, right?

src/Update/AnimatedGifWindowWinForms.cs
paulcbetts: This probably needs some cleanup, I kinda yolo'd out this code

src/Update/TaskbarManager.cs
shiftkey: [Vista SP2](https://docs.microsoft.com/en-us/dotnet/framework/get-started/system-requirements)

paulcbetts: 👎 👎 👎

PR #1076, opened 2017-06-26
Add capability for StubExecute.exe to display a splash image

StephenMcConnel:
This can be useful for C#/.Net programs which are notoriously slow to
start up in order to reassure users that something is happening.

### DIFF
--- a/src/StubExecutable/SplashImageUtils.cpp
+++ b/src/StubExecutable/SplashImageUtils.cpp
@@ -1,0 +1,334 @@
+// SplashImageUtils.cpp : methods for displaying a splash image while starting
+// the real program.  Many of these methods are adapted from those found at
+// http://faithlife.codes/blog/2008/09/displaying_a_splash_screen_with_c_introduction/
+// The original code is Copyright 2007-2008 Logos Bible Software and licensed
+// under an MIT style license according  to
+// http://faithlife.codes/blog/2008/09/logos_code_blog_license/.  The revisions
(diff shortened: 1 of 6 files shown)


hatton:
Specifically, @StephenMcConnel did this for the initial "cold start" scenario right after a Windows restart. A cold start of our app was clocked at 27 seconds before .net UI code could run, on one slow computer. With this PR, they see a splash screen immediately and then when the .net code is ready to show its own, it close down this initial splash screen.

lukeskinner:
We have our own fork of squirrel which allows a custom stub to be set during releasify. We use it to handle users who close the application mid update and to fallback to reinstalling the software on failure (anti virus removing the actual application). 

The actual code changes to allow a custom stub exe is pretty minor. Is this something you would accept in a PR?

Users can then create with their own stubs - adding splash images etc. 

The only issues I've found with this approach is the taskbar shortcut. If the user has just pinned the shortcut it will bypass the stub until the next update. In my app I just rewrite the taskbar shortcut at startup if it's not using the stub exe.

hatton:
> The only issue I've found with this approach is the taskbar shortcut. If the user has just pinned the shortcut it will bypass the stub until the next update. In my app I just rewrite the taskbar shortcut at startup if it's not using the stub exe.

@lukeskinner thanks for pointing that out. Is your code that rewrites the taskbar shortcut available somewhere?

StephenMcConnel:
This isn't going to be accepted, so I'm closing the issue.

Issue #630, opened 2016-03-16
A simple way to package an app


poma:
I'm not sure why documentation recommends things like OctoPack and Auto.Squirrel when packaging can be easily integrated directly into build process using only nuget and squirrel. Just define a build target in .csproj file:

``` xml
<Target Name="AfterBuild" Condition=" '$(Configuration)' == 'Release'">
  <GetAssemblyIdentity AssemblyFiles="$(TargetPath)">
    <Output TaskParameter="Assemblies" ItemName="myAssemblyInfo"/>
  </GetAssemblyIdentity>
  <Exec Command="nuget pack MyApp.nuspec -Version %(myAssemblyInfo.Version) -Properties Configuration=Release -OutputDirectory $(OutDir) -BasePath $(OutDir)" />
  <Exec Command="squirrel --releasify $(OutDir)MyApp.%(myAssemblyInfo.Version).nupkg" />
</Target>
```

this will generate a nuget package from .nuspec file setting version from AssemblyInfo.cs and place it in OutDir (by default `bin\Release`). Then it will generate release files from it.

Here is an example .nuget file for myApp:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
  <metadata>
    <id>MyApp</id>
    <!-- version will be replaced by MSBuild -->
    <version>0.0.0.0</version>
    <title>title</title>
    <authors>authors</authors>
    <description>description</description>
    <requireLicenseAcceptance>false</requireLicenseAcceptance>
    <copyright>Copyright 2016</copyright>
    <dependencies />
  </metadata>
  <files>
    <file src="*.*" target="lib\net45\" exclude="*.pdb;*.nupkg;*.vshost.*"/>
  </files>
</package>
```

Solution needs to have `nuget.exe` available for example by installing `NuGet.CommandLine` package. And also it suffers from this bug when sometimes NuGet packages are not loaded properly and throws `nuget/squirrel is not recogized (9009)` errors.

> you may simply need to restart Visual Studio so the Package Manager Console will have loaded all the package tools

Other than that I think it's the simplest solution that fits most simple apps and should be recommended in Getting Started tutorial. I'm not sure whether including `nuget.exe` in squirrel package is a good practice but it could eliminate an additional dependency on `NuGet.CommandLine` and make MSBuild integration work out of the box.

kenbailey:
[...]

PR #1085, opened 2017-07-02
Updated the documentation for Visual Studio Build Packaging

lukeskinner:
Minor change to the code example for passing the version to squirrel. Before it would pass the full version i.e. 1.0.0.0.

Since squirrel v1.5.0 this is not valid as it enforces SemVer. 

This fix was taken from #630.


paulcbetts:
:+1:

PR #1087, opened 2017-07-04
Fix for invalid bsdiff file being created in delta package

Keboo:
When the bsdiff fails to create a file it leaves a blank .bsdiff file sitting in the delta package which will cause the updates to fail. This occurs because the [BinaryPatchUtility assumes](https://github.com/Squirrel/Squirrel.Windows/blob/master/src/Squirrel/BinaryPatchUtility.cs#L299) that it should always be able to read at least 32 bytes from the stream.

If you would also like a fix to the BinaryPathcUtility.Apply method to handle empty streams I am happy to create that as well, however I think this should be sufficient.

Will likely submit another PR after I track down why the creation of the delta packages is throwing a divide by zero exception on my pdb file.

Squirrel.Windows NuGet version: 1.7.6

### DIFF
--- a/src/Squirrel/DeltaPackage.cs
+++ b/src/Squirrel/DeltaPackage.cs
@@ -207,22 +207,23 @@ namespace Squirrel
                     this.Log().Warn("We couldn't create a delta for {0}, attempting to create bsdiff", targetFile.Name);
                 }
             }
-
-            var of = default(FileStream);
+            
             try {
-                of = File.Create(targetFile.FullName + ".bsdiff");
-                BinaryPatchUtility.Create(oldData, newData, of);
+                using (FileStream of = File.Create(targetFile.FullName + ".bsdiff"))
+                {
+                    BinaryPatchUtility.Create(oldData, newData, of);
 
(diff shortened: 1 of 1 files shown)


src/Squirrel/DeltaPackage.cs
```diff
@@ -207,22 +207,23 @@ void createDeltaForSingleFile(FileInfo targetFile, DirectoryInfo workingDirector
...
-                of = File.Create(targetFile.FullName + ".bsdiff");
-                BinaryPatchUtility.Create(oldData, newData, of);
-
-                // NB: Create a dummy corrupt .diff file so that older 
-                // versions which don't understand bsdiff will fail out
-                // until they get upgraded, instead of seeing the missing
-                // file and just removing it.
-                File.WriteAllText(targetFile.FullName + ".diff", "1");
+                using (FileStream of = File.Create(targetFile.FullName + ".bsdiff"))
+                {
```
paulcbetts: Move bracket ⬆️

paulcbetts:
Merged with fixups in ffaf5f2392d34a415e433581fffe78c4ff96862b, thanks!

PR #1088, opened 2017-07-05
Fixing issue where exception stack trace was getting smashed.

Keboo:
Using `throw ex;` will cause the [original stack trace to be overwritten](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/throw). This change properly wraps the exception so the original stack trace is properly preserved inside the inner exception.


paulcbetts:
:+1:, thanks for the fix!

PR #1090, opened 2017-07-10
Pass through silentInstall from ApplyReleases

guyfletch:
Pass silentInstall from ApplyReleases to ApplyReleasesImpl.ApplyReleases, instead of setting it explicitly to false.

Optional argument defaulting to false, so this shouldn't break any existing code.

### DIFF
--- a/UpdateManager.cs
+++ b/UpdateManager.cs
@@ -1,0 +1,317 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.Contracts;
+using System.Drawing;
+using System.Globalization;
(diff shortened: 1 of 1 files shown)


shiftkey:
@guyfletch the added file in this pull request lives at the root of the repository, but the source file lives at [`src/Squirrel/UpdateManager.cs`](https://github.com/Squirrel/Squirrel.Windows/blob/master/src/Squirrel/UpdateManager.cs). 

Can you update the pull request to update the correct file so the diff of changes is clearer? Because the file is not in the right location there will be no functional change when merging this PR in it's current form...

PR #1112, opened 2017-07-26
Readme: Recursive clone instead of submodules init

judge2020:
The --[recursive](https://git-scm.com/docs/git-clone#git-clone---recurse-submodulesltpathspec) option when cloning is the equivalent of doing a `git submodule update --init --recursive`.


paulcbetts:
Into it! Thanks :)

PR #1115, opened 2017-07-27
provide a user agent when fetching the releases entry

shiftkey:
Making it consistent with [this usage](https://github.com/Squirrel/Squirrel.Windows/blob/14d6dc1599676a9d83daba60e73507d12b73164d/src/Squirrel/UpdateManager.Factory.cs#L68).


paulcbetts:
:+1:

PR #1106, opened 2017-07-21
Validate update package signature

arul-gupta:
I've added an option to validate the packages before applying the update.
If signature verification is requested, it will continue with update only if the update package is signed and the Subject property of the certificate matches the one from the current executable.
If current app is not signed, it will skip the validation check even if user requests signature validation.

This fix was needed to restrict app updates to only trusted sources.

PR #1121, opened 2017-08-04
Verify that the update is coming from a trusted source

arul-gupta:
I've added an option to validate the packages before applying the update using [catalog file](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/catalog-files).

**Steps in verification**
If the user passes the verifySignature flag as true to CheckForUpdates function, the following things will be checked:
- Download RELEASES.cat from the same update url location.
- Verify that the catalog file downloaded is signed using a valid certificate issued by a trusted authority.
- Verify that the update comes from a trusted source by checking the CN of current app certificate and catalog file certificate.
- Ensure that the checksum in RELEASES.cat file matches the calculated checksum of downloaded RELEASES file.
- If any of the above steps fail, throw an exception else continue as usual.

**Changes to Squirrel packaging**
- Realisify option now creates a RELEASES.cat along with all the other files. This is a windows catalog file containing the SHA1 checksum of RELEASES file. 
- RELEASES.cat is then signed if signing options have been provided.
- This catalog file should be put at the same location as other files at upload url.

PS: I've closed PR #1106 which intended to achieve the same result by signing the .nupkg file but that is supposedly not supported with signtool.


athomson1:
Anyone here can push this up? We're also needing this :)

Trinitek:
I was exploring the possibility of wrapping the nupkgs in a signed cabinet, but this is a much less intrusive solution. +1

paulcbetts:
Hi, I don't think we're going to go down this route, aside from the problems in the PR itself (doesn't follow the style, catalog files can't be generated on Mac/Linux) it adds significant complexity. I would instead pursue something like HTTPS Cert Pinning to verify authenticity of the download (I'd take a PR for that in fact)

Issue #1143, opened 2017-09-05
Suppressing Install .net framework modal dialog


turgayselvi:
I know when we use -s (quite) option, squirrel setup doesn't show Install Model dialog but also it doesn't show any progress about installation.

My case is, I want to skip .net installation dialog but I want to show .net installation progress. Is there any chance to add another parameter option suppress/ignore model dialog.

fr4gles:
The only way I see it will be possible is by using [Chaining the .NET Framework installation to your app's setup with Chaining by Using a Custom UI](https://docs.microsoft.com/en-us/dotnet/framework/deployment/deployment-guide-for-developers#chaining)

Useful links:

- https://docs.microsoft.com/en-us/dotnet/framework/deployment/deployment-guide-for-developers
- https://docs.microsoft.com/en-us/dotnet/framework/deployment/deployment-guide-for-developers#chaining

turgayselvi:
@fr4gles  Thanks for links but my issue isn't related framework itself.

In Squirrel.Windows/src/Setup/FxHelper.cpp

Setup.exe is skipping this dialog is isQuiet is true. But if isQuiet is true, it doesn't show any progress bar for .net installation. Because installation is being started with quite mode.

```
`if (!isQuiet) {
		CTaskDialog dlg;
		TASKDIALOG_BUTTON buttons[] = {
			{ 1, L"Install", },
			{ 2, L"Cancel", },
		};


		dlg.SetButtons(buttons, 2);
		dlg.SetMainInstructionText(GetInstallerMainInstructionForVersion(version));
		dlg.SetContentText(GetInstallerContentForVersion(version));
		dlg.SetMainIcon(TD_INFORMATION_ICON);

		dlg.SetExpandedInformationText(GetInstallerExpandedInfoForVersion(version));

		int nButton;
		if (FAILED(dlg.DoModal(::GetActiveWindow(), &nButton)) || nButton != 1) {
			return S_FALSE;
		}
	}`
```

My suggestion is to add another parameter

```
`if (!isQuiet || !skipNetModalDialog) {
		CTaskDialog dlg;
		TASKDIALOG_BUTTON buttons[] = {
			{ 1, L"Install", },
			{ 2, L"Cancel", },
		};


		dlg.SetButtons(buttons, 2);
		dlg.SetMainInstructionText(GetInstallerMainInstructionForVersion(version));
		dlg.SetContentText(GetInstallerContentForVersion(version));
[...]

PR #1151, opened 2017-09-12
Skip net modal dialog

turgayselvi:
I added --skipNetModalDialog parameter for Setup.exe as alternative for -s (quite) mode.

When user run setup.exe, we don't want to ask any dialog but we want to show .net framework installation progress.  If we use -s parameter, Setup.exe won't ask dialog but won't show any progress. It will install silently.


fr4gles:
I guess it resolves https://github.com/Squirrel/Squirrel.Windows/issues/1143 too, am I right?

turgayselvi:
@fr4gles 

Yes, It resolves #1143

turgayselvi:
I had to replace --skipNetModalDialog parameter with --doNotAskInstallDotNet .  
--skipNetModalDialog includes '-s'  which means silent mode. Setup.exe thinks this is silent mode. I would fix the other way but I decided to change the parameter name that I defined before.

fpadilha:
Nice PR! =)

PR #1164, opened 2017-10-02
Update log messages to have a more professional tone

kblum:
This pull requests updates the copy of some messages to have a more professional tone.

Since the Squirrel log file can be viewed by end users, it should not contain anything that can be perceived as negative by customers of Squirrel-based applications.


paulcbetts:
You're certainly welcome to keep this in your own fork, but I don't think we need to update these

PR #1171, opened 2017-10-09
Feature/localizeinstaller

teh173:


paulcbetts:
Soooooo here's a question - why didn't you just display the license dialog as the first screen of your app? The idea of Squirrel is that you move these kinds of things into the app, because it's way easier to write and test them, instead of moving it into the installer

teh173:
Oops, I meant that PR for our fork. Though I'm happy to share.

Believe me... I wanted to do it the easy way... but... lawyers... :)

paulcbetts:
No worries, Squirrel.Windows is open-source so that people _can_ use it as a jumping off point to build what they want!

PR #1120, opened 2017-08-03
Update docs for MSBuild

poma:
This is the latest version that includes all suggestions from #630

Changes: 

* Automatically searches for nuget.exe and squirrel.exe and doesn't require opening package console
* Displays well formatted errors if executables are not found
* Changed MyApp to $(TargetName) to match project name bu default
* Added quotes around paths to support spaces in file path
* Note about new PackageReference directive

PR #1188, opened 2017-10-31
Squirrel Docs Update - Visual Studio Packaging (includes PR #1120)

kenbailey:
@paulcbetts, when  you get a chance can you merge this in. Includes a bunch of good changes to the Visual Studio Packaging docs by @poma, @Yitzchok, @markfox1, @PHeonix25, @awbacker, and others. 

It includes a little bit of formatting I've done.

Thanks!

### DIFF
--- a/docs/using/visual-studio-packaging.md
+++ b/docs/using/visual-studio-packaging.md
@@ -10,23 +10,28 @@ Squirrel packaging can be easily integrated directly into your build process usi
 The first step is to define a build target in your `.csproj` file.
 
 ```xml
-<ItemGroup>
-  <NuGetCommandLine Include="..\packages\NuGet.CommandLine.*\tools\nuget.exe">
-    <InProject>False</InProject>
-  </NuGetCommandLine>
-  <Squirrel Include="..\packages\Squirrel.Windows.*\tools\squirrel.exe">
-    <InProject>False</InProject>
-  </Squirrel>
-</ItemGroup>
 <Target Name="AfterBuild" Condition=" '$(Configuration)' == 'Release'">
-  <!-- Add some nice errors for the next person that comes along -->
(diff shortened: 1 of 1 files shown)


docs/using/visual-studio-packaging.md
```diff
@@ -60,13 +84,12 @@ Here is an example `MyApp.nuspec` file for the above build target example. Don't
...
-
-```pm
-PM>  Install-Package NuGet.CommandLine
-```
+Please be aware of the following when using this solution:
 
-Note: this will not work if your solution uses new `PackageReference` directive instead of `packages.config` file for NuGet dependencies because it will be unable to find nuget.exe and squirrel.exe. In that case change path to executables accordingly.
+* It suffers from a bug when sometimes NuGet packages are not loaded properly and throws nuget/squirrel is not recogized (9009) errors. **Tip:** In this case you may simply need to restart Visual Studio so the Package Manager Console will have loaded all the package tools
+* If you get a `'squirrel' is not recognized as an internal or external command` error, you need to add the full path to `Squirrel.exe` in the build target `Exec Command` call.  
+* You can create a separate build configuration for building a package if you wish to avoid running `--releasify` every time you build your `Release` configuration (e.g., create a new `Package` configuration based on `Release`).
```
poma: Those 3 points no longer apply when using macros and `Error` tags, you should delete them

docs/using/visual-studio-packaging.md
```diff
@@ -60,13 +84,12 @@ Here is an example `MyApp.nuspec` file for the above build target example. Don't
...
+* This will not work if your solution uses new `PackageReference` directive instead of `packages.config` file for NuGet dependencies because it will be unable to find nuget.exe and squirrel.exe. In that case change path to executables accordingly.
```
poma: This probably can be moved to the relevant section of `Build Target Configuration`

poma:
In my opinion this description is unnecessarily long, especially describing what every msbuild tag does. Leaving PR #1120 open so that maintainer can decide whether to merge additional text or not.

kenbailey:
@poma, fair enough. Thanks for the feedback.

PR #1206, opened 2017-11-18
Add criteria for using deltas to update docs.

relsqui:
I came here looking for this and didn't find it. :) The rationale I'm adding is based on this line/function, let me know if I'm interpreting it incorrectly:

https://github.com/Squirrel/Squirrel.Windows/blob/f63edb85d29ed6df9a3735690139442c87ebca01/src/Squirrel/UpdateInfo.cs#L76


paulcbetts:
👍

PR #1233, opened 2018-01-11
Go back to WPF instead of WinForms

damieng:
This PR effectively reverts #1061 to go back to WPF.  While the WinForms one was more efficient with larger GIF animations it suffered the following problems;

1. Splash not draggable (easily solved)
2. High dpi support (not supported in WinForms)
3. Decent alpha (not supported in WinForms)

The WpfAnimatedGif component released an update that reduces the memory somewhat.  It's still not enough to effectively use 150 frame animated gifs but is okay for smaller ones such as what Atom uses.

I'm not sure what to do about the 2x scaling change.  Recent versions of Squirrel require you double your splash screen size as a way to pseudo-support high dpi.  That was a breaking change for people and not having it here is another breaking change.

It would be nice if we had some kind of background-x.gif support for different dpi's but I'm not sure that's easily integrated.  Nor is it easy for us to detect if somebody has doubled their gif size or not so there really is no automatic way to make all people happy here.

A simple option might be to look for background-2x.gif and display that dpi-scaled and if it doesn't exist look for background.gif and use as-is.


src/Update/Program.cs
```diff
@@ -60,13 +60,13 @@ int executeCommandLine(string[] args)
         {
             var animatedGifWindowToken = new CancellationTokenSource();
 
+#if !MONO
             // Uncomment to test Gifs
-            /*
```
paulcbetts: We probably want to bring this back

src/Update/Program.cs
```diff
@@ -148,7 +148,7 @@ int executeCommandLine(string[] args)
 #if !MONO
                 case UpdateAction.Install:
                     var progressSource = new ProgressSource();
-                    if (!silentInstall) {
```
paulcbetts: Whitespace only changes whyyyyyyyy

paulcbetts:
Well that was easy.

src/Update/Program.cs
damieng: image

src/Update/Program.cs
damieng: Oops.  Well at least you know I was testing it.

src/Update/Program.cs
paulcbetts: You don't actually need to go and change these back, I just wanted to whinge about it :P

Issue #1146, opened 2017-09-06
Taskbar icon during setup is the default .net icon


o11ie:
If _setup.exe_ is the output from the squirrel **releasify** command.

The _setup.exe_ file has been customised to include the desired icon.

However, while setup runs, the taskbar icon is the default .net winforms icon.

Is there a way to prevent the taskbar icon appearing or replace the icon with a custom icon?

kjpacheco:
The icon in the taskbar is set in your code. 
In your solution, right click on your solution>Properties>Application.  In there, you will see a block for 'Resources' where you can specify your icon image.

fr4gles:
o11ie is right. I faced same behaviour and it does nothing in common with application icon couse it is ok. Just setup icon during install is default winforms app icon.

kjpacheco:
Sorry, I guess I misunderstood which icons we were talking about.
I haven't figured out how to change the setup.exe icon. For my project, I don't have a need to alter it from the default.  Once everything is installed, custom icon for my application.exe is there

o11ie:
Thanks for the comments. I'm still having this issue though. Will I have to dig into the Squirrel source code to fix this? Functionally everything is perfect. It's just that professional touch that is missing.

teh173:
I also see this issue. It looks like animated GIF window is showing this icon even though I don't have a GIF setup.

o11ie:
Hi thanks for everyone's comments corroberating my issue. Does anyone know of a fix or work around, yet? Do these sorts of issue ever get added to the backlog?

jonoomph:
Hello! Any updates on this? ... or known workarounds?

Tobias-B-Besemer:
I found the following issue in the latest Atom and Git installer. Guess is (still) the same issue. Using Win7.

![2019-11-11 21_12_33-](https://user-images.githubusercontent.com/11881073/68686589-22ff8780-056c-11ea-8f9b-6db255e40c34.jpg)

![2019-11-08 03_31_07-](https://user-images.githubusercontent.com/11881073/68686608-2b57c280-056c-11ea-95cf-961e9b33df69.jpg)

PR #1242, opened 2018-01-26
set the already sepcified icon to AnimatedGifWindow, too

szanto90balazs:
The icon, specified during releasify is not visible at all in the task bar, while installing. The default icon is visible. Attempting to fix https://github.com/Squirrel/Squirrel.Windows/issues/1146

image

This PR aims to fix it by bundling the icon file, just like the loading gif and setting the `AnimatedGifWindow`'s Icon property.


jonoomph:
Any possibility we can get this PR merged? This affects all Squirrel installers, and would be great to fix.

paulcbetts:
👍 , thanks!

PR #1246, opened 2018-01-31
Pass the full path of the installer executable through to app on first-run

### LINKED ISSUE #1240: Passing the setup exe file name through to app
Is there any way to hook in to retrieve the filename that setup.exe was renamed to? e.g. when receiving the `--squirrel-install` in the app, does it get passed through at all to anywhere?

petemill:
Setup.exe passes the path of the installer executable to Update.exe via `--installer-path` flag.

Update.exe passes the path to the installed app on first-run via the `--squirrel-installer-path` flag.

Fix #1240

### DIFF
--- a/src/Setup/UpdateRunner.cpp
+++ b/src/Setup/UpdateRunner.cpp
@@ -259,8 +259,12 @@ gotADir:
 		lpCommandLine = L"";
 	}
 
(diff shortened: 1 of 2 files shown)


ElAdrian05-git (approved):
Rusv

PR #1278, opened 2018-03-22
Synchronous code signing

jonoomph:
Switching the Releasify method to code sign files in a synchronous method (as opposed to async). For larger projects, with more than 130 signable files, the async handling of this causes lots of deadlocks and freezes in our docker build image.

This simplified, synchronous version works awesome, has no freezes or deadlocks, and in our testing (with lots of signable files) actually completes faster than the previous async version. Go figure. =)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -181,4 +181,3 @@ tests/Test.xml
 ## CPP db crap
 *.db
 *.opendb
-/.vs/Squirrel/v15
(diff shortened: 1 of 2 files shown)


jonoomph:
Needs another small adjustment first

paulcbetts:
Just push more commits, you don't need to close this PR

paulcbetts:
Also, you can really just set the concurrency factor of ForEachAsync to `1`, and also, please match the style

PR #1295, opened 2018-04-24
Strong passwords (anything except [a-zA-Z0-9_]) are now being obfuscated correctly for signtool calls

pfeigl:
Currently passwords when using signtool.exe are obfuscated using an Regex with \w+ which is only "word characters".

What we actually want is any "non whitespace character", which is \S+.

This is beeing fixed by this PR.


HenrikHoyer (approved):

anaisbetts:
@HenrikHoyer Please do not approve PRs on repos that you do not maintain.

PR #1306, opened 2018-05-01
Add .NET Framework 4.7.2 support.

peters:


Foda:
This is causing our installer to fail on newer versions of Windows (insider) so hopefully this will be merged soon. The fall 2018 update will ship with 4.7.2 support built-in, which seems like it will cause all existing Squirrel apps to fail the .Net version check (unless I'm misunderstanding how `FxHelper` works.

paulcbetts:
@Foda Hi, can you send in a SquirrelSetup.log where you see this failure?

Foda:
@paulcbetts Of course. We'll ask for one from the users seeing this and I'll create a new issue and link it here once we get the log. Ignore my previous post -- I was misunderstanding the version check.

cbenard:
@paulcbetts is there any way we can get the 4.7.2 changeset merged in anyway?

paulcbetts:
@cbenard Yeah, I'm going to do a new release tonight, I'll get this in

riverar:
This is a band-aid for a regression in Windows 10 RS5 17755 and higher. We have to push Microsoft to fix this (missing NDP registry key) as it's near RTM.

paulcbetts:
@riverar You Know What To Do™. Surprised this hasn't showed up on AppCompat yet, that missing reg key would break Many Many Things

PR #1328, opened 2018-06-01
Works around locked file signing

TonyValenti:
Sometimes Explorer and other apps will get a lock on the DLL or EXE which will prevent SignTool from running and will then throw off the entire signing and packaging process.

This adjusts the code so that if signing fails, it will re-attempt three times before throwing the error.  Also, the output error has been adjusted to show the return code and the output from SignTool that way you can know more than just "signing failed".

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -643,17 +643,25 @@ namespace Squirrel.Update
                 if (!File.Exists(exe)) exe = "signtool.exe";
             }
 
-            var processResult = await Utility.InvokeProcessAsync(exe,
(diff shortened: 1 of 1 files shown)


bzuillsmith (commented):
I know this is quite old, but I've been getting the problem lately. Oddly, your commit appears to change every line of this file while it seems you just added a for loop of 3 attempts in one spot. Maybe line endings are the issue here? Additionally I wonder if it would be beneficial to add a delay in your loop of attempts just to give a little more time for the explorer or whatever process to release it. I dunno how fast the async call runs.

robmen:
Would be very interesting to review when the change doesn't modify the entire file.

PR #1340, opened 2018-06-19
Fixed grouping of app icons in Windows Taskbar when creating shortcuts.

MathiasWasAlreadyTaken:
This change is for when using the --createShortcut option.
Changed the appUserModelId to not have the full path to the exe, instead just point out the application's name.
The same id should then be used at startup in your app using the Shell32.lib function SetCurrentProcessExplicitAppUserModelID, which will make the icons in the taskbar look like it's the same application.
Ex: SetCurrentProcessExplicitAppUserModelID("com.squirrel.[nuspec_id].[exe_name]");

By doing this the user can pin either the bootstrap or the actual application to the taskbar, and they will end up in the same group in the taskbar.

### DIFF
--- a/src/Squirrel/UpdateManager.ApplyReleases.cs
+++ b/src/Squirrel/UpdateManager.ApplyReleases.cs
@@ -163,7 +163,7 @@ namespace Squirrel
                     if (!locations.HasFlag(f)) continue;
 
                     var file = linkTargetForVersionInfo(f, zf, fileVerInfo);
-                    var appUserModelId = String.Format("com.squirrel.{0}.{1}", zf.Id.Replace(" ", ""), exeName.Replace(".exe", "").Replace(" ", ""));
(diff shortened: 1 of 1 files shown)


bddckr:
I don't think this PR is necessary. Instead I believe the issue you're trying to solve stems from wrong usage of the method: The parameter name `exeName` implies that you should pass the executable's **name, not the path to it**.

With that in mind the existing code looks correct to me, but of course one could argue that the usage of `Path.GetFileNameWithoutExtension` is "cleaner" than doing `.Replace(".exe", "")`.

Does this explanation fix the issue you're trying to solve with this PR @MathiasWasAlreadyTaken?

MathiasWasAlreadyTaken:
Thanks for the input. You're right, I will change my code to just the exe name.
We should perhaps update the documentation about this, and how to use SetCurrentProcessExplicitAppUserModelID. There are several applications out there that doesn't use this.

PR #1355, opened 2018-07-24
Update: Ensure we don't create stub for generated stub executables

IonMiron:


src/Squirrel.nuspec
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
-    <version>1.8.0</version>
+    <version>1.8.1</version>
```
paulcbetts: Don't change version numbers in PRs

src/Update/Program.cs
```diff
@@ -392,13 +392,15 @@ public void Releasify(string package, string targetDir = null, string packagesDi
...
                 var rp = new ReleasePackage(file.FullName);
                 rp.CreateReleasePackage(Path.Combine(di.FullName, rp.SuggestedReleaseFileName), packagesDir, contentsPostProcessHook: pkgPath => {
-                    new DirectoryInfo(pkgPath).GetAllFilesRecursively()
-                        .Where(x => x.Name.ToLowerInvariant().EndsWith(".exe"))
-                        .Where(x => !x.Name.ToLowerInvariant().Contains("squirrel.exe"))
-                        .Where(x => Utility.IsFileTopLevelInPackage(x.FullName, pkgPath))
-                        .Where(x => Utility.ExecutableUsesWin32Subsystem(x.FullName))
-                        .ForEachAsync(x => createExecutableStubForExe(x.FullName))
-                        .Wait();
+                    var exeFiles = new DirectoryInfo(pkgPath)
```
paulcbetts: Can we not change the style here?

src/Squirrel/Utility.cs
```diff
@@ -74,7 +74,14 @@ public static IEnumerable<FileInfo> GetAllFilesRecursively(this DirectoryInfo ro
         {
             Contract.Requires(rootPath != null);
 
-            return rootPath.EnumerateFiles("*", SearchOption.AllDirectories);
+            return rootPath.GetAllFilesRecursively("*");
+        }
+
+        public static IEnumerable<FileInfo> GetAllFilesRecursively(this DirectoryInfo rootPath, string searchPattern)
```
paulcbetts: Why is this a function now? It does the same thing

paulcbetts:
What does this do? Can you fill in the description describing what problem this is trying to solve?

src/Squirrel/Utility.cs
IonMiron: The idea is to have an overload that takes the search pattern as a parameter (to be used when collectiong exe files for stub creation)

IonMiron:
Hey @paulcbetts ,
What do you think about the PR? Can you have another look?

PR #1361, opened 2018-08-02
Fix div-by-zero errors on files which don't have any changed content

peppy:
There are cases where the length in write operations becomes zero, at which point this code will throw an exception and fail to create bsdiff deltas.

As this is critical to our production workflow, it has been temporarily published as [ppy.Squirrel.windows](https://www.nuget.org/packages/ppy.squirrel.windows/) (along with several other PRs that are not merged into official releases yet).

Test files which cause the issue to reproduce (large): https://www.dropbox.com/sh/p13oiqir02xzny1/AABUHNZeP2u2T-AjViFYCG6na?dl=0

### DIFF
--- a/src/Squirrel/BinaryPatchUtility.cs
+++ b/src/Squirrel/BinaryPatchUtility.cs
@@ -230,10 +230,13 @@ namespace Squirrel.Bsdiff
             WriteInt64(controlEndPosition - startPosition - c_headerSize, header, 8);
 
             // write compressed diff data
-            using (WrappingStream wrappingStream = new WrappingStream(output, Ownership.None))
-            using (var bz2Stream = new BZip2Stream(wrappingStream, CompressionMode.Compress))
+            if (dblen > 0)
             {
-                bz2Stream.Write(db, 0, dblen);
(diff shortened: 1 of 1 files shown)


caioproiete:
@peppy The test files are no longer available on your Dropbox. Would you be able to upload them somewhere else?

[image](https://www.dropbox.com/sh/p13oiqir02xzny1/AABUHNZeP2u2T-AjViFYCG6na?dl=0)

adabru:
@peppy, does https://github.com/Squirrel/Squirrel.Windows/pull/1421 achieve the same thing?

peppy:
I have no idea, but the test files are still availble if you'd like to try.

We have long since given up hope and maintain our own fork.

adabru:
Thanks for open sourcing and distributing your fork on nuget. I switched to your fork.

MichaelBelousov:
looks related to #1784

caesay:
This is the same problem that was fixed in #1421 and has already been merged to master. I don't know if this fix has been released on nuget, but this PR can be closed. It is also fixed in [my fork](https://github.com/clowd/Clowd.Squirrel) which has superseded the `ppy.Squirrel.windows` package.

PR #1365, opened 2018-08-03
Expose an $(SquirrelToolsPath) property for nuget consumers

### LINKED ISSUE #1364: Expose an $(SquirrelToolsPath) property for nuget consumers
When using the nuget package, it's often convenient to
be able to execute it in some targets. However, since the nuget
install path depends on the version of the package, it's typically
harder than necessary to figure out the path to tools.

kzu:
When using the nuget package, it's often convenient to
be able to execute it in some targets. However, since the nuget
install path depends on the version of the package, it's typically
harder than necessary to figure out the path to tools.

This adds a .props file that is automatically imported by nuget
when installed in a project, therefore making it trivial to invoke
the tools by just using the new $(SquirrelToolsPath) MSBuild property
from anywhere in the project or its imported targets.

Since it's in a .props, it means that even downstream packages that
add a dependency on this package can leverage the tool too without
having to know what version of the Squirrel.Windows package is in use.

Fixes #1364

### DIFF
--- a/src/Squirrel.nuspec
+++ b/src/Squirrel.nuspec
@@ -25,6 +25,7 @@
     <file src="Squirrel\bin\Release\Squirrel.*" target="lib\Net45" />
     <file src="Squirrel\bin\Release\NuGet.Squirrel.*" target="lib\Net45" />
     <file src="Squirrel\bin\Release\ICSharpCode.*" target="lib\Net45" />
+    <file src="Squirrel.props" target="build" />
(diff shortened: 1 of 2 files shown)


kzu:
Does it ship straight from CI to nuget.org?

paulcbetts:
> Does it ship straight from CI to nuget.org?

kzu:
Bummer :P

PR #1368, opened 2018-08-06
Added updateOnly option to createShortcut command

SamvelRaja:
The createShorcutForExecutable method has the ability to do differentiation between whether it is a updateOnly action or a create Action but this ability is not exposed via command line. 

This PR enables the feature, feel free to decline it if it is not aligning with the roadmap.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -99,6 +99,7 @@ namespace Squirrel.Update
                 bool shouldWait = false;
                 bool noMsi = (Environment.OSVersion.Platform != PlatformID.Win32NT);        // NB: WiX doesn't work under Mono / Wine
                 bool noDelta = false;
+                bool updateOnly = false;
 
                 opts = new OptionSet() {
                     "Usage: Squirrel.exe command [OPTS]",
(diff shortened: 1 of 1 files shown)


src/Update/Program.cs
```diff
@@ -132,6 +133,7 @@ int executeCommandLine(string[] args)
                     { "l=|shortcut-locations=", "Comma-separated string of shortcut locations, e.g. 'Desktop,StartMenu'", v => shortcutArgs = v},
                     { "no-msi", "Don't generate an MSI package", v => noMsi = true},
                     { "no-delta", "Don't generate delta packages to save time", v => noDelta = true},
+                    { "updateOnly", "Argument that will be used while creating shortcut, decides whether to just update the shortcuts", v => updateOnly = false},
```
kutu:
```suggestion
                    { "updateOnly", "Argument that will be used while creating shortcut, decides whether to just update the shortcuts", v => updateOnly = true},
```
if option present, `updateOnly` variable should be true, not false

SamvelRaja:
@kutu Changes looks fine?

kutu:
@SamvelRaja yes, i have tested it, and it works as expected, thx

SamvelRaja:
@kutu  will it be merged?
A lot of electron apps would be in need of this feature.

kutu:
How do I know, I'm nobody in this project
But I use this in my project now, thanks for PR

vidia (approved):
Thank you for making this change. I hope that this can get merged in because I have need for a similar feature in a project I'm working on. It looks valid to me, but I am still new to this code so I'll defer to the maintainers. 

Looking over the contributing docs, it looks like this needs to be moved to "develop" instead of master. I'm happy to re-create this PR myself if needed to keep this alive. 

Adding @shiftkey who marked this as unmergable. What can I do to pick up this work and get it moving?

PR #1370, opened 2018-08-13
Build setup.exe in temp and copy to destination

easel:
Prevents users from reading a partially written setup.exe.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -388,16 +388,16 @@ namespace Squirrel.Update
 
             ReleaseEntry.WriteReleaseFile(releaseEntries, releaseFilePath);
 
-            var targetSetupExe = Path.Combine(di.FullName, "Setup.exe");
+            var tempTargetSetupExe = Path.GetTempFileName(); 
             var newestFullRelease = releaseEntries.MaxBy(x => x.Version).Where(x => !x.IsDelta).First();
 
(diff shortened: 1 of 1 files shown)


src/Update/Program.cs
```diff
@@ -458,16 +458,26 @@ public void Releasify(string package, string targetDir = null, string packagesDi
...
+            var targetSetupExe = Path.Combine(di.FullName, "Setup.exe");
+            if(File.Exists(targetSetupExe))
+            {
+                File.Delete(targetSetupExe);
             }
+            File.Move(tempTargetSetupExe, targetSetupExe);
 
-            if (generateMsi) {
+            if (generateMsi)
+            {
```
paulcbetts: y tho

easel:
Meh. Sorry about that. Totally blaming visual studio for that one. Tightened up the other trailing braces as well.

paulcbetts:
@easel Can you tell me more about this fix? What is it trying to solve?

easel:
It's trying to solve the problem of building setup files directly to a published location. With the current code, if clients try to run the installer while the build is running, they get an invalid file. With this fix, the file is generated offline and replaced atomically so clients never see an invalid file.

robmen:
The problem with using %TMP% is that virus checkers often get huffy about writing temporary .exes then moving them around. It'd be better if we standardize on an `obj` folder that is relative to the build process (just like VS does for projects).

PR #1401, opened 2018-10-24
Fix #1283 (Spaces in filename cause ReleaseEntry Regex to fail)

### LINKED ISSUE #1283: Spaces in filename cause ReleaseEntry Regex to fail
I'm in a situation where Squirrel updating fails when iterating over one particular release entry

`E6651B768DB82076E7113086558548B4DA4F8F6F Artifact - BoxHeader - e7ee795c-99e1-4ecb-a72c-9a41461d8198.xml.shasum 8669`

using this Regex:

`^([0-9a-fA-F]{40})\s+(\S+)\s+(\d+)[\r]*$`

The culprit appears to be that `(\S+)` will only match non-whitespace characters. I've played a bit around with the Regex Tester at http://regexstorm.net/tester, and my suggestion would be to change the Regex to something similar to

`^([0-9a-fA-F]{40})\s+(.*).shasum[\s+](\d+)[\r]*$`

We don't really want to be forced to change the filenames or escape the spaces, so I'm interested in a solution where we can avoid that. Thanks in advance.

cbenard:
Regex change allows spaces in the filename for while requiring at least one non-space character to fix #1283. Worked with @JohnThomson to refined the regex in the issue comments.

### DIFF
--- a/src/Squirrel/ReleaseEntry.cs
+++ b/src/Squirrel/ReleaseEntry.cs
@@ -92,7 +92,7 @@ namespace Squirrel
             return zp.IconUrl;
         }
 
-        static readonly Regex entryRegex = new Regex(@"^([0-9a-fA-F]{40})\s+(\S+)\s+(\d+)[\r]*$");
+        static readonly Regex entryRegex = new Regex(@"^([0-9a-fA-F]{40})\s+(\S[\S ]*?)\s+(\d+)[\r]*$");
         static readonly Regex commentRegex = new Regex(@"\s*#.*$");
         static readonly Regex stagingRegex = new Regex(@"#\s+(\d{1,3})%$");
         public static ReleaseEntry ParseReleaseEntry(string entry)


JohnThomson:
There are extensive unit tests for this function...could you figure out a new one that fails without the fix?

cbenard:
@JohnThomson, I didn't have a dev env set up for this. I went ahead and cloned in the NuGet submodule and opened it in VS2017. Even in master, there are a ton of failing tests for me. Also, I looked at the tests for `UpdateManager` and it's not pleasant. It's actually copying files out of the `test\fixtures` folder and there aren't any that have spaces.

I don't want to try to add files with spaces to the repo and given that there are failing tests already, this is beyond the time I have to spend on it. I'm confident in the RegEx that we worked on together. If you'd like to mess with the tests, please do your own PR. If the tests were all green already, I would have likely proceeded.

image

JohnThomson:
I think I've seen problems running all the tests, too. But there's a suite of tests, the second last entry in your screen shot, for ReleaseEntryTests, and many of them test this specific method, passing it a string (the file contents). So I believe it should be fairly easy to add a test that requires the fix without adding any files.
I'd be willing to do that if Paul wants it...not just in case he might, though.

PR #1426, opened 2018-12-12
GitHub require TLS 1.2

wolkesson:
Apps using .NET framework 4.5 will fail when checking for updates. Don't know if this is the correct place, but I spend a day figuring this out...


paulcbetts:
@wolkesson thanks a ton for sending in the PR after you figured this out and saving other developers some suffering!

PR #1432, opened 2019-01-02
 Add Logging to Releasify

Horb:
# Add Logging to Releasify

The following refactor was brought about because we (Age Partnership) were running into difficulty using Squirrel in the Azure DevOps Build Pipelines and existing logging wasn't sufficient for our needs.

I don't expect any functionality to have changed.

The Releasify method in Update.exe has been broken up into;

* ensureConsole
* ValidateBaseUrl
* ValidatePaths
* PrepareFiles
* ProcessFiles
* CleanupFiles
* WriteReleaseFile
* CreateSetupExe
* GenerateMsi

Breaking Releasify up into smaller methods allows stack traces to be more useful when identifying issues.

The refactoring was done using ReSharper. Only significant change is the use of `out var` declarations and passing variables using `ref`. I expect a more significant refactor could make this cleaner.  Debug logs have been added to the start of each method and the whole Releasify method is now in a "try / catch / log / throw" block.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -62,6 +62,7 @@ namespace Squirrel.Update
         string bootstrapperExe = default(string);
         string backgroundGif = default(string);
         string baseUrl = default(string);
+        string signingParameters = default(string);
         string setupIcon = default(string);
         string frameworkVersion = "net45";
         bool noMsi = (Environment.OSVersion.Platform != PlatformID.Win32NT);        // NB: WiX doesn't work under Mono / Wine
@@ -98,7 +99,6 @@ namespace Squirrel.Update
                 bool shouldWait = false;
                 string releaseDir = default(string);
                 string icon = default(string);
-                string signingParameters = default(string);
                 string processStartArgs = default(string);
                 string shortcutArgs = default(string);
 
@@ -423,7 +423,6 @@ namespace Squirrel.Update
             }
         }
 
-        string signingOpts;
         void ProcessFiles()
         {
             this.Log().Debug("Processing Files");
@@ -440,7 +439,7 @@ namespace Squirrel.Update
                         .ForEachAsync(x => createExecutableStubForExe(x.FullName))
                         .Wait();
 
-                    if (signingOpts == null) return;
+                    if (signingParameters == null) return;
 
                     new DirectoryInfo(pkgPath).GetAllFilesRecursively()
                         .Where(x => Utility.FileIsLikelyPEImage(x.Name))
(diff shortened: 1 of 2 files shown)


src/Update/Program.cs
```diff
@@ -345,8 +345,27 @@ public async Task Uninstall(string appName = null)
...
+                ReleasifyCreateSetupExe(package, bootstrapperExe, backgroundGif, signingOpts, setupIcon, frameworkVersion, di, releaseEntries, out var targetSetupExe);
+                ReleasifyGenerateMsi(package, signingOpts, generateMsi, targetSetupExe);
+                this.Log().Debug("Completed Releasify");
+            } catch (Exception ex) {
+                this.Log().Error(ex);
+                throw;
+            }
+        }
 
+        private void ReleasifyValidateBaseUrl(ref string baseUrl)
```
paulcbetts: Remove all of the `private` keywords here

src/Update/Program.cs
```diff
@@ -345,8 +345,27 @@ public async Task Uninstall(string appName = null)
 
         public void Releasify(string package, string targetDir = null, string packagesDir = null, string bootstrapperExe = null, string backgroundGif = null, string signingOpts = null, string baseUrl = null, string setupIcon = null, bool generateMsi = true, string frameworkVersion = null, bool generateDeltas = true)
         {
-            ensureConsole();
+            try {
+                this.Log().Debug("Begin Releasify");
+                ensureConsole();
+                ReleasifyValidateBaseUrl(ref baseUrl);
```
paulcbetts: I like the concept of reorganizing like this, but we don't need to prefix all of the function names with `Releasify`, just use the rest

src/Update/Program.cs
```diff
@@ -345,8 +345,27 @@ public async Task Uninstall(string appName = null)
...
+                ReleasifyValidatePaths(ref targetDir, ref packagesDir, ref bootstrapperExe);
+                ReleasifyPrepareFiles(package, targetDir, out var di, out var toProcess, out var processed, out var releaseFilePath, out var previousReleases);
```
paulcbetts: Many of these parameters are used in nearly every function (`package`, `targetDir`) - you probably want to make them instance variables instead - having 8+ parameters to a function with most of them being `ref`s and `out`s doesn't make the code Better, it only makes it more difficult to change

paulcbetts:
I appreciate the concept but I'm not sure we've made the code Better here. So many variables are shared between all of these that R#'s result makes changing the code frustrating. An alternate approach would be to move many of these inner variables to instance fields, then try the R# Extract Method trick again

Horb:
Refactor to introduce instance variables is complete. I need to test to make sure I haven't broken anything. Once I've tested I'll report back here.

Horb:
@paulcbetts I think this is good to go now. Some variable names are no longer used and the name from command line parsing is used instead such as; signingOpts/signingParameters, package/target.

PR #1434, opened 2019-01-02
 Use AutoFlush = true in Logger for Update.exe

Horb:
# Use AutoFlush = true in Logger for Update.exe

Using AutoFlush = true allows;

* Logs to still be written in a catastrophic failure condition.
* User to watch the log in real time using something like `Get-Content -Wait`
or `tail -f`

From my testing and due to the nature of the operations conducted by Update.exe the performance impact of AutoFlush is negligible.

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -857,7 +857,7 @@ namespace Squirrel.Update
     class SetupLogLogger : Splat.ILogger, IDisposable
     {
         TextWriter inner;
-        readonly object gate = 42;
+        readonly object gate = new object();
         public Splat.LogLevel Level { get; set; }
 
         public SetupLogLogger(bool saveInTemp)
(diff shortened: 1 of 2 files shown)


src/Update/Program.cs
```diff
@@ -857,7 +857,7 @@ public void Raise(int i)
     class SetupLogLogger : Splat.ILogger, IDisposable
     {
         TextWriter inner;
-        readonly object gate = 42;
+        readonly object gate = new object();
```
paulcbetts: This isn't any different

src/Update/Update.csproj
```diff
@@ -39,13 +39,13 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="DeltaCompressionDotNet">
-      <HintPath>..\..\packages\DeltaCompressionDotNet.1.1.0\lib\net45\DeltaCompressionDotNet.dll</HintPath>
+      <HintPath>..\..\packages\DeltaCompressionDotNet.1.1.0\lib\net20\DeltaCompressionDotNet.dll</HintPath>
```
paulcbetts: Revert this change please

paulcbetts:
I'm Into this! Thanks for the contribution

paulcbetts:
✨

PR #1440, opened 2019-01-21
fix nuget build files not being included automatically in consumer projects

domyd:
This relates to #1365 and #1366.

I've got a simple `netstandard2.0` library that references `squirrel.windows` 1.9.0 and prints the `SquirrelToolsPath` property to the build console:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="squirrel.windows" Version="1.9.0" />
  </ItemGroup>

  <Target Name="PrintSquirrelToolsPath" AfterTargets="PostBuildEvent">
    <Message Importance="High" Text="SquirrelToolsPath is: $(SquirrelToolsPath)" />
  </Target>

</Project>
```

However, the output of that is:
```
1>Target PrintSquirrelToolsPath:
1>  SquirrelToolsPath is:
```
(I also tried it with a new WPF `net45` application project with the old `packages.config` format, and it yielded the same result.)

Looking at the full build output, the `Squirrel.props` and `Squirrel.targets` files are never imported by MSBuild. The NuGet docs mention that [the .props and .targets file names need to match the package ID](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package) in order for them to automatically be included by MSBuild. And indeed, if I rename the build files to `squirrel.windows.props` and `squirrel.windows.targets`, it works:
```
1>Target PrintSquirrelToolsPath:
1>  SquirrelToolsPath is: C:\Users\Dominik.Mydlil\.nuget\packages\squirrel.windows\1.9.1-custom.1\tools
```

So that's what this PR does - it renames the existing `Squirrel.props` and `Squirrel.targets` to `squirrel.windows.props` and `squirrel.windows.targets` so that the auto-include works.


paulcbetts:
Great fix!

PR #1442, opened 2019-01-25
Allow for `app-0.0.0` to be correctly identified

scottrippey:
For local builds of my app, we hard-code the version number to `0.0.0`.
After updating to the latest Squirrel, we found that even after installing, the app was no longer launchable.

This change ought to fix that behavior, and allow `app-0.0.0` to be launchable again.  And since all other filenames ought to be unique, it should have no other side effects.

### DIFF
--- a/src/StubExecutable/StubExecutable.cpp
+++ b/src/StubExecutable/StubExecutable.cpp
@@ -67,7 +67,7 @@ std::wstring FindLatestAppDir()
 
 		version::Semver200_version thisVer(s);
 
-		if (thisVer > acc) {
+		if (thisVer >= acc) {
 			acc = thisVer;
 			acc_s = appVer;
 		}
--- a/src/StubExecutable/StubExecutable.cpp
+++ b/src/StubExecutable/StubExecutable.cpp
@@ -53,7 +53,7 @@ std::wstring FindLatestAppDir()
 		return NULL;
 	}
 
-	version::Semver200_version acc("0.0.0");
+	version::Semver200_version acc("0.0.0-0");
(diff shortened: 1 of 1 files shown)


paulcbetts:
Did you actually test this? I don't see how it would fix the bug, you still end up in the check after the end of the loop. Also like, maybe just don't do this?

scottrippey:
I didn't test, and you're right, it was not going to work.  I changed the approach. I haven't tested this either, but I do think it's a better approach.  I'd be glad to add tests, if you think this has potential.

Regarding "maybe just don't do this" ... that'll be my workaround -- to use version `0.0.1` for local builds.  However, a few counter-arguments ....
- When building, Squirrel happily creates the `*-0.0.0 (Setup).exe` installer.  Unfortunately, though, the app only launches once (post-install), and can never be launched again.  So this does seem like a bug.
- Version `0.0.0` worked in Squirrel 1.6 (this bug only started when upgrading to 1.9)
- Version `0.0.0` works fine on Mac (when building Electron apps)
- Version `0.0.0` is the most logical version number for local builds (vs. `0.0.1` etc). We have several places in our app that detect local builds via `0.0.0` (which could obviously change), but an upstream fix would be ideal.

paulcbetts:
I'm not sure that I want to fix this. I would accept a PR that would fail builds if 0.0.0 is used.

scottrippey:
Fair enough, thanks for taking the time to consider this!

scottrippey:
Just to follow up, I successfully worked around this issue by changing the version number to `0.0.1`.

PR #1447, opened 2019-02-14
Fallback to package.Id if title not available

james1465:
If a package title is not set in the nuspec tile, normal user installs work fine, but the installation folder set in the Wix for the machine wide installer becomes "%programfiles%\ Installer\ .... " - note the space where the title should be. 

When the folder is created this space is silently removed by the operating system, however in the registry Run entry, the space is present. As the name in the registry and the actual path don't match, the machine wide installer never runs. 

This change means that the installer falls back to using the id if the name is not available

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -732,7 +732,7 @@ namespace Squirrel.Update
             var templateText = File.ReadAllText(Path.Combine(pathToWix, "template.wxs"));
             var templateData = new Dictionary<string, string> {
                 { "Id", package.Id },
-                { "Title", package.Title },
+                { "Title", package.Title ?? package.Id },
                 { "Author", company },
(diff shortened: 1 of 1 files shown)


paulcbetts:
I would rather that Releasify simply fails if title isn't present, can you do this instead?

james1465:
Yes I can do that if you're sure, the MS documentation on .nuspec files does this about the title though:

> title
> A human-friendly title of the package, typically used in UI displays as on nuget.org and the Package Manager in Visual Studio. If not specified, the package ID is used.
[Microsoft](https://docs.microsoft.com/en-us/nuget/reference/nuspec)

So it does feel like going against the spirit of the nuspec format a little not to fallback to the ID. 

It's your project through - let me know your thoughts and I'll implement it.

PR #1453, opened 2019-02-28
update support, 64bit support, better naming

bitdisaster:
This PR brings several MSI improvements that were repeatedly requested by customers (IT administrators)

**Update Support**
The WIX template already accommodated a `<UpgradeCode/>` tag that is necessary uninstalling older versions of the MSI. However, to actually trigger an uninstall when installing a new one, it was missing the `<MajorUpgrade/>` tag. No more. Old version will be removed now.

**64bit Support**
When offering your App as 64-bit, administrators also expect the installer package to behave 64-bit. Meaning that the installer is puts things under `Program Files` not `Program Files (x86)` and registry keys are written under `HKLM\SOFTWARE\...` and not `HKLM\SOFTWARE\WOW6432Node\...`  A new optional parameter `-arch=VALUE` allows to define the architecture for the packaging process. While only the value `x64` will currently have an effect, it leaves the door open for future architecture differentiations.

**Better Naming**
Many customer struggle with the concept what the MSI in Squirrel is and what machine-wide installer in this context means. Naming the MSI `MyApp Machine-Wide installer` implies that actually MyApp gets installed machine-wide when in reality only an installer that installs MyApp gets installed. So this PR attempts to make it more clear what the MSI is and what it installs. The best way to describe it is an installer for a tool that helps deploy `MyApp` on per-user bases. Therefor the MSI will be listed as `MyApp Per-User Deployment`. Also the new folder naming should make it clear what gets installed. `C:\Program Files\MyApp Deployment\MyAppDeploymentTool.exe`

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -100,6 +100,7 @@ namespace Squirrel.Update
                 bool shouldWait = false;
                 bool noMsi = (Environment.OSVersion.Platform != PlatformID.Win32NT);        // NB: WiX doesn't work under Mono / Wine
                 bool noDelta = false;
+                bool packageAs64Bit = false;
 
                 opts = new OptionSet() {
                     "Usage: Squirrel.exe command [OPTS]",
@@ -134,6 +135,7 @@ namespace Squirrel.Update
                     { "no-msi", "Don't generate an MSI package", v => noMsi = true},
                     { "no-delta", "Don't generate delta packages to save time", v => noDelta = true},
                     { "framework-version=", "Set the required .NET framework version, e.g. net461", v => frameworkVersion = v },
+                    { "msi-win64", "Mark the MSI as 64-bit, which is useful in Enterprise deployment scenarios", _ => packageAs64Bit = true},
                 };
 
                 opts.Parse(args);
@@ -184,7 +186,7 @@ namespace Squirrel.Update
                     break;
 #endif
                 case UpdateAction.Releasify:
-                    Releasify(target, releaseDir, packagesDir, bootstrapperExe, backgroundGif, signingParameters, baseUrl, setupIcon, !noMsi, frameworkVersion, !noDelta);
+                    Releasify(target, releaseDir, packagesDir, bootstrapperExe, backgroundGif, signingParameters, baseUrl, setupIcon, !noMsi, packageAs64Bit, frameworkVersion, !noDelta);
                     break;
                 }
             }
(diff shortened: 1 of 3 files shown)


src/Update/Program.cs
```diff
@@ -134,6 +135,7 @@ int executeCommandLine(string[] args)
                     { "no-msi", "Don't generate an MSI package", v => noMsi = true},
                     { "no-delta", "Don't generate delta packages to save time", v => noDelta = true},
                     { "framework-version=", "Set the required .NET framework version, e.g. net461", v => frameworkVersion = v },
+                    { "arch=", "Defines the architecure of the MSI package, e.g x64", v => packageAs64Bit = v == "x64"},
```
paulcbetts: This is kind of a fib, since I can't specify like, `arm` or `ia64` here. Can we rename this to `msi-win64=true`?

vendor/wix/template.wxs
```diff
@@ -1,8 +1,9 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension" xmlns:netfx="http://schemas.microsoft.com/wix/NetFxExtension">
-  <Product Id="*" Name="{{Title}} Machine-Wide Installer" Language="1033" Codepage="{{Codepage}}" Version="!(bind.FileVersion.{{Id}}.exe)" UpgradeCode="{{IdAsGuid1}}" Manufacturer="{{Author}}">
+  <Product Id="*" Name="{{Title}} Per-User Deployment" Language="1033" Codepage="{{Codepage}}" Version="!(bind.FileVersion.{{Id}}.exe)" UpgradeCode="{{IdAsGuid1}}" Manufacturer="{{Author}}">
```
paulcbetts: How about instead of "Per-User", we write "{{Title}} Automatic Deployment" - having the name "Per-User" in the install entry makes people think, "Oh yeah, that's the version installed per-user".

src/Update/Program.cs
bitdisaster: My thought behind using the more generic option here was to have the freedom for possible future adjustments related to architecture without introducing yet another option. But if you feel `
msi-win64` is the better choice then it works for me to.

vendor/wix/template.wxs
bitdisaster: Yeah I was hoping the "Deployment"-part of the name would make this clear. Automatic work for me to.

bitdisaster: What about 
{{Title}} Deployment Tool"

bitdisaster:
all addressed

paulcbetts:
Thanks for the PR!

PR #1456, opened 2019-03-13
Logging improvements

bitdisaster:
This PR brings various improvements around logging with the main purpose of making them easier to understand for developer or CE engineers that do not have knowledge of the internals of Squirrel.

**1.) Bugfix because of multiple process logging to the same file**
With exemption of the `--uninstall` case logs get written to the location of the executing binary. During install and update scenarios the `Update.exe` in `%localappdate%\MyApp` can run in multiple process at the same time due to squirrel hooks. Previous code has tried to avoid collision but since the file was opened with ReadWrite access a collision would never happen. Instead partial or all content from one or the other process was missing. While the collision code is fixed and remains, a new naming strategy will also avoid collision (see 2.)
**2.) Better Squirrel log naming**
Squirrel logs leaves logs in multiple locations depending on the executing binary's location. Furthermore, log file can contain information from multiples runs of different Squirrel commands such `--update`, `--createShortcut`, etc. To make it easier to understand the content log files are now named by the invoked Squirrel command, e,g `Squirrel-Update.log`, `Squirrel-Uninstall.log`. The naming also avoids collision described in 1.)
**3.) No errors/ exceptions in happy path**
Even if an installation or update went perfectly, the logs contained errors/exceptions. Without deeper knowledge of Squirrel its not easy to understand whether this is of any concern. From now on the logs will be nice and shiny if Squirrel walked the happy path.
**4.) Clear end of logging session**
When reading the log files its not clear whether a command such as `--install` was executed successful. It seems like a rather sudden stop. A final log entry at the end adds some peace of mind.
**5.) Add log level**
While Squirrel uses log level internally they were not written to the log files. Now we will know if an entry is just and info or a  warning or an error.

### DIFF
--- a/src/Squirrel/UpdateManager.ApplyReleases.cs
+++ b/src/Squirrel/UpdateManager.ApplyReleases.cs
@@ -536,25 +536,25 @@ namespace Squirrel
             // directory are "dead" (i.e. already uninstalled, but not deleted), and
             // we blow them away. This is to make sure that we don't attempt to run
             // an uninstaller on an already-uninstalled version.
-            async Task cleanDeadVersions(SemanticVersion originalVersion, SemanticVersion currentVersion, bool forceUninstall = false)
+            async Task cleanDeadVersions(SemanticVersion currentVersion, SemanticVersion newVersion, bool forceUninstall = false)
             {
-                if (currentVersion == null) return;
+                if (newVersion == null) return;
 
                 var di = new DirectoryInfo(rootAppDirectory);
                 if (!di.Exists) return;
 
-                this.Log().Info("cleanDeadVersions: for version {0}", currentVersion);
-
-                string originalVersionFolder = null;
-                if (originalVersion != null) {
-                    originalVersionFolder = getDirectoryForRelease(originalVersion).Name;
-                    this.Log().Info("cleanDeadVersions: exclude folder {0}", originalVersionFolder);
-                }
+                this.Log().Info("cleanDeadVersions: checking for version {0}", newVersion);
 
                 string currentVersionFolder = null;
                 if (currentVersion != null) {
                     currentVersionFolder = getDirectoryForRelease(currentVersion).Name;
-                    this.Log().Info("cleanDeadVersions: exclude folder {0}", currentVersionFolder);
+                    this.Log().Info("cleanDeadVersions: exclude current version folder {0}", currentVersionFolder);
+                }
+
+                string newVersionFolder = null;
+                if (newVersion != null) {
+                    newVersionFolder = getDirectoryForRelease(newVersion).Name;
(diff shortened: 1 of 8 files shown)


src/Squirrel/UpdateManager.ApplyReleases.cs
```diff
@@ -536,25 +536,25 @@ internal void unshimOurselves()
             // directory are "dead" (i.e. already uninstalled, but not deleted), and
             // we blow them away. This is to make sure that we don't attempt to run
             // an uninstaller on an already-uninstalled version.
-            async Task cleanDeadVersions(SemanticVersion originalVersion, SemanticVersion currentVersion, bool forceUninstall = false)
+            async Task cleanDeadVersions(SemanticVersion currentVersion, SemanticVersion newVersion, bool forceUninstall = false)
```
paulcbetts: This is a much better name, 👍

src/Squirrel/UpdaterIntention.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Squirrel
```
paulcbetts: Squirrel doesn't use one-class-per-file, can you put this in IUpdateManager.cs (or maybe a different file, whatever makes Sense)

src/Update/StartupOption.cs
```diff
@@ -0,0 +1,85 @@
+﻿using Mono.Options;
```
paulcbetts: Into this refactor 👍

paulcbetts:
Can you rebase against master to get rid of the weird merge commits?

src/Squirrel/UpdaterIntention.cs
bitdisaster: sure thing

bitdisaster:
Yeah, those commits refuse to go away. You are not on Electron HQ anymore

bitdisaster:
Got it done. All feedback addressed so far.

paulcbetts:
@bitdisaster Wait one more thing, this PR breaks a public interface, can we find a way to do this without the breaking change to `CheckForUpdate`?

bitdisaster:
🤔

bitdisaster:
I can't think of way to get totally rid of it because at the place where I need it, I have no other context on whether its an update or install scenario. However, we could make it virtually a non breaking change by moving the parameter to the end and apply a default value. `Task<UpdateInfo> CheckForUpdate( bool ignoreDeltaUpdates = false, Action<int> progress = null, UpdaterIntention intention = UpdaterIntention .Update);`

paulcbetts:
> However, we could make it virtually a non breaking change by moving the parameter to the end and apply a default value

Works for me, we don't promise ABI compatibility

paulcbetts:
👍

Issue #1462, opened 2019-03-26
Incompatible with Splat 7.0.1


ondrasvoboda:
The latest Splat version 7.0.1 has some breaking changes that makes squirrel updates failing with:
Method not found: 'Void Splat.IFullLogger.Info(System.String, !!0)'

So avoid updating of Splat package, until this version is supported by squirrel.

paulcbetts:
Fixed in #1464

jeremy-bridges:
Is this the permanent fix for this issue? As other dependencies of our application depend on Splat 7.0.1 and newer, Squirrel will prevent us from moving forward with other updates. Akavache 6.5.1 is an example. I wouldn't mind submitting the pull request to use 7.0.1. Sound good?

jeremy-bridges:
I've discovered this update would require an update to .NET Framework 4.6.1. Likely, you would judge this update as a breaking change, moving Squirrel to 2.0.0. For us, this is not an issue as this is already a requirement of our apps. How many other users need to stay with 4.5?

Thieum:
@jeremy-bridges considering .NET 4.5 is no long supported [since january 2016](https://devblogs.microsoft.com/dotnet/support-ending-for-the-net-framework-4-4-5-and-4-5-1/), I'd consider it legitimate to upgrade to a newer version. I'm not fully aware of all the implications. We choose to discuss the different dependency upgrades in [this issue](https://github.com/Squirrel/Squirrel.Windows/issues/897).

jeremy-bridges:
I've commented there about moving to 4.6.1. 

Another part of this pull request will have to be how to implement newly required parts of Splat's iLogger interface. There's three new methods on that interface that seem to be implemented to select an underlying logger for any particular provider:

void Write(Exception exception, string message, LogLevel logLevel);
void Write(string message, Type type, LogLevel logLevel);
void Write(Exception exception, string message, Type type, LogLevel logLevel);

I'll need to provide an implementation for these. I can certainly provide an implementation of the first one. It will simply add the exception to the string being written to the log. I assume Squirrel's setup logger won't need a real implementation for the other two overloads (that take a type). The purpose of these is also not very obvious to me, looking through Splat's repository. An example of NLog's implementation can be found below. Seem alright to throw a not-implemented exception for these overloads?

https://github.com/reactiveui/splat/blob/master/src/Splat.NLog/NLogLogger.cs

Thieum:
[...]

PR #1464, opened 2019-03-27
Update Squirrel.nuspec

rschoenbach:
Restricting splat version to prevent incompatibility.

Tested Squirrel with Splat 6.1.7 and everything worked correctly. 

Helps prevent against #1462


paulcbetts:
Thanks @rschoenbach

PR #1482, opened 2019-05-07
Add editorconfig

### LINKED ISSUE #1478: add a simple editorconfig to the project

gojanpaolo:
Fixes #1478

@shiftkey @Thieum I added the .editorconfig generated by IntelliCode to have a starting point. I think it might be a better idea to start from this instead of manually extracting it from the project as it has a lot of inconsistencies.

Unfortunately, IntelliCode can only generate C# code style. My next task is to add C++.

We can then [format all files](https://marketplace.visualstudio.com/items?itemName=munyabe.FormatAllFiles) after the initial .editorconfig is complete.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -86,11 +86,11 @@ dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
 
 #Style - qualification options
 
-#prefer events not to be prefaced with this. or Me. in Visual Basic
+#prefer events not to be prefaced with this.
 dotnet_style_qualification_for_event = false:suggestion
-#prefer fields not to be prefaced with this. or Me. in Visual Basic
+#prefer fields not to be prefaced with this.
 dotnet_style_qualification_for_field = false:suggestion
-#prefer methods not to be prefaced with this. or Me. in Visual Basic
+#prefer methods not to be prefaced with this.
 dotnet_style_qualification_for_method = false:suggestion
-#prefer properties not to be prefaced with this. or Me. in Visual Basic
+#prefer properties not to be prefaced with this.
 dotnet_style_qualification_for_property = false:suggestion
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,96 @@
+[*.cs]
+
+#Core editorconfig formatting - indentation
+
+#use soft tabs (spaces) for indentation
+indent_style = space
+
+#Formatting - indentation options
+
+#indent switch case contents.
+csharp_indent_case_contents = true
+#indent switch labels
+csharp_indent_switch_labels = true
+
+#Formatting - new line options
+
+#place catch statements on a new line
+csharp_new_line_before_catch = true
+#place else statements on a new line
+csharp_new_line_before_else = true
+#require finally statements to be on a new line after the closing brace
+csharp_new_line_before_finally = true
+#require braces to be on a new line for properties, object_collection_array_initializers, types, control_blocks, lambdas, and methods (also known as "Allman" style)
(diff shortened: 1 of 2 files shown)


Thieum:
Should we consider and apply the style to the Core project as it's some sort of external dependencies? Parsing the code to see for the review, it became clear that this Core has a really different style than the Squirrel and Tests projects.

amaitland:
An exclusion can be added to ignore the `vendor` folder, see https://stackoverflow.com/questions/30310396/possible-to-ignore-exclude-file-folder-from-editorconfig

`root = true` and a `Default Settings` section should be added. The main https://editorconfig.org/ has an example (as I'm sure everyone is aware). 

They also have a list of fairly prominent projects using `.editorconfig` which is a great reference at https://github.com/editorconfig/editorconfig/wiki/Projects-Using-EditorConfig

https://github.com/dotnet/corefx/blob/master/.editorconfig
https://github.com/dotnet/roslyn/blob/master/.editorconfig
https://github.com/aspnet/AspNetCore/blob/master/.editorconfig

gojanpaolo:
@amaitland added root = true and default settings. :) I think we should apply a different .editorconfig on the NuGet.Core project as @Thieum suggested so we have a code style in it too in case the code needs to be updated in the future.

btw, it looks like C++ has limited support for .editorconfig (first-time dealing with C++ here). I was trying to find like `new_line_before_open_brace` but there isn't any. I believe the default settings should cover most of C++ style and then override some like `indent_style = tab`.

Thieum:
@gojanpaolo for C++, we could go the clangformat way which is supported by VS as documented here: https://devblogs.microsoft.com/cppblog/clangformat-support-in-visual-studio-2017-15-7-preview-1/

gojanpaolo:
Ah! that's why I don't get anything when I googled "C++ EditorConfig". good to know. thanks! Will study and add it later tonight :)

gojanpaolo:
@Thieum @amaitland it looks like NuGet.Core is a git submodule and points to https://github.com/paulcbetts/NuGet . So updating it is technically out of scope of this repository thus we won't have a separate editorconfig for it as previously discussed.

gojanpaolo:
@Thieum I tried generating using the Squirrel solution excluding NuGet.Core and C++ projects and it matches the latest in this PR. I think we may benefit in the long term having a single code style for all C# squirrel project. :)

.editorconfig
```diff
@@ -0,0 +1,83 @@
...
+# Language keyword and framework type
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+
+# Qualification
+dotnet_style_qualification_for_event = false:suggestion
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+
+[/vendor/nuget/**.cs]
```
shiftkey: @gojanpaolo I think the rules below also need to handle `cpp` files inside `vendor/nuget/`. When I ran [`eclint fix **/*.cpp`](https://github.com/jedmao/eclint) to test out what this would detect it triggered a lot of whitespace changes. 

Would that be possible to add in?

gojanpaolo: @shiftkey We can by doing this `[/vendor/nuget/**.{cs,cpp}]`
but I don't see `cpp` files inside `vendor/nuget/`

I ran `eclint fix **/*.cpp` and the whitespace changes were on files inside the `src` folder and it's because of the `default settings` in the `.editorconfig`

PR #1483, opened 2019-05-07
docs: Add .NET Application Settings to FAQ

### LINKED ISSUE #198: Inconsistent config file after update.
I update my app with Squirrel and I call `RestartApp()`.

Since the old installation folder is left intact and a new one is created inside the `rootAppDirectory`, the config file is written from scratch. As a result, any settings retained in the old installation's config file, don't migrate to the new config file. This is a problem if the app makes use of `Properties.Settings` to keep application settings.

I should either create some logic myself to port the settings I want to the new config file, or dump the Application settings and use a custom settings file, probably saved in `AppData\MyAppData`.

Am I missing something?

gojanpaolo:
Fixes #198

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -66,7 +66,7 @@ var isSquirrelInstall = File.Exists(updateDotExe);
 4. **My previous application version is still around after the update. Doesn't Squirrel clean up old versions?**  
    The current and immediately previous version of your application are not deleted on clean up (see [issue #589](https://github.com/Squirrel/Squirrel.Windows/issues/589)). 
 5. **How can I persist the [.NET Application Settings](https://docs.microsoft.com/en-us/dotnet/framework/winforms/advanced/application-settings-overview) after an update?**
-   See (https://github.com/Squirrel/Squirrel.Windows/issues/198#issuecomment-299262613) for a simple workaround if you want to keep using .NET Application Settings. Another solution is to use a different app/user settings mechanism such as storing it in JSON file.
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -65,6 +65,8 @@ var isSquirrelInstall = File.Exists(updateDotExe);
    You can't. So, how can you do it? The basic trick that ClickOnce uses is, you have a folder of EXEs and DLLs, and an Application Shortcut. When ClickOnce goes to update its stuff, it builds a completely *new* folder of binaries, then the last thing it does is rewrite the app shortcut to point to the new folder.
 4. **My previous application version is still around after the update. Doesn't Squirrel clean up old versions?**  
    The current and immediately previous version of your application are not deleted on clean up (see [issue #589](https://github.com/Squirrel/Squirrel.Windows/issues/589)). 
+5. **How can I persist the [.NET Application Settings](https://docs.microsoft.com/en-us/dotnet/framework/winforms/advanced/application-settings-overview) after an update?**
+   See (https://github.com/Squirrel/Squirrel.Windows/issues/198#issuecomment-299262613) for a simple workaround if you want to keep using .NET Application Settings. Another solution is to use a different app/user settings mechanism such as storing it in JSON file.
```
shiftkey: Tweaking the last sentence to focus more on _where_ the settings are stored, rather than what they're stored as...

```suggestion
   See (https://github.com/Squirrel/Squirrel.Windows/issues/198#issuecomment-299262613) for a simple workaround if you want to keep using .NET Application Settings. Alternatively, consider using a solution that lets you control where the settings are persisted, and store settings in an app-specific location under `%LOCALAPPDATA%`.
```

Issue #964, opened 2017-02-28
Squirrel version/rollback solutions


markgould:
Didn't know where to put this but I thought I'd share a solution I came up with that addresses some of the headaches of rollback/pre-release tags in versions

We use Octopus Deploy to deploy some of our software products to customers servers (and then invoke squirrel on the server) and this works great with a couple of caveats.  The main pain points are pre-release packages and rolling back. Both of these are solved by just creating a new Octopus release with new version each time, but this makes it hard to track on the dashboard what version any given customer might be on.

The absurdly simple solution I came up with today, is to simply increment the last release number. It's a little overkill but I quickly hacked this together:

```
function Get-NextVersionNumber {
    # parameters
    param ([string] $ReleasePath)

    $ReleaseFile = [System.IO.Path]::Combine($ReleasePath, "RELEASES")

    if ([System.IO.File]::Exists($ReleaseFile) -eq $false) {
        return "0.0.1"
    }

    $Lines = [System.IO.File]::ReadAllLines($ReleaseFile);

    if ($Lines.Count -eq 0) {
        return "0.0.1"
    }

    $LastLine = $Lines[$Lines.Count-1]

    if ($LastLine -cmatch '^([0-9a-fA-F]{40})\s+(\S+)\s+(\d+)[\r]*$') {
	    $result = $matches[2] -creplace '(-full|-delta)?\.nupkg$', ''
        if ($result -cmatch '\d+(\.\d+){0,3}(-[A-Za-z][0-9A-Za-z-]*)?') {
            $split = $matches[0].Split(".");            
            $split[$split.Count-1] = ($split[$split.Count-1] -as [int])+1
            return [string]::Join(".", $split)
        }
    }

    return "0.0.1";     
}
```

I can then call Get-NextVersionNumber with my releases path and in my nuget pack command, simply override the version with this.

If I need to rollback a version, I simply re-deploy the old release in Octopus and it will increment the release version. This also solves the pre-release issue on the version, as we no longer use the Octopus release number for the squirrel release. 

Example:

| Octopus Action | Squirrel Release           | 
| ------------- |-------------|
| Deploy 1.0.0      | 0.0.1|
| Deploy 1.1.0   | 0.0.2|  
| Re-deploy 1.0.0 | 0.0.3|   
| Deploy 1.2.0-alpha.1   | 0.0.4|  

Hope this helps someone!

wjrogers:
[...]

Issue #796, opened 2016-08-12
How can the user move to a previous version?


SissiKonsta:
Hello,

Is it possible to give the user the ability to move to a previous version?

Thank you in advance.

kneuvirth:
I have been trying to do this same thing. Any help would be appreciated.

andymartinwork:
Have them uninstall the current version, and send them the installer for the old version?

gojanpaolo:
@shiftkey duplicate of other rollback issues (#524, #964)

Thieum:
@shiftkey @gojanpaolo the "original" downgrade issue is #145 Should we consolidate?

gojanpaolo:
@Thieum yes. I think we should update the documentation and then close the two I mentioned above as those discusses the workarounds/strategies instead of having it built-in the library.

shiftkey:
Closing this out in favour of #524

Issue #524, opened 2015-12-15
Rollback strategy


grandtiger:
What's the recommended strategy for rollback? Ideally, rollback should be achieve by rollback the folder / URL that contains all the binaries (Setup.msi, Setup.exe, RELEASES, xxxx-delta-nupkg, xxxx-full.nupkg) to the previous state. Is this possible? If not, is there any plan to introduce support for rollback?

paulcbetts:
What do you mean by 'rollback' - do you mean from a developer perspective? Or from a user failed-install perspective?

bengotow:
This may not be related, but I was just going to file a similar ticket: I've noticed that we're unable to get Squirrel.Windows to move to older versions. eg: if the app has the version number `0.3.32` and we send it a RELEASES file with the following content:

```
C1589D46DB2C1BECF4363DD14ED0C2A4C5420A0A nylas-0.3.29-full.nupkg 86917771
```

It downloads the nupkg, unpacks it, and then deletes it, relaunching the pre-existing version. We're able to roll back to previous releases with Squirrel.Mac (which I don't think parses the version numbers). Would be really great if the same thing worked on Windows!

paulcbetts:
@bengotow I might try to support it but it's this massive special-case everywhere (i.e. delta logic gets weird, etc etc). Easier thing at the moment is to just create a rollback package that has a higher version #

trumbjd:
The reason I would push for this to be added is being able to create "channels" for Test, Dev, Prod without having separate installs for each.

paulcbetts:
@trumbjd You can totally do this already via just changing the URL. All you need to do is make sure your version number is consistent between the different channels (i.e. always increasing)

trumbjd:
That is super.  I must have missed this when I did an eval last year.  Appreciate the correction!

raybooysen:
This is pretty important for my project.  If we roll-out a buggy version, it would be nice to easily roll-back to a previous package by modifying the RELEASES file instead of rolling forward.  

Rolling forward is generally a longer process (builds, etc.) and for a buggy release, moving quickly is better for our clients.

fleischman718:
Same here. 
This is a killer feature for me, I'm still stuck with default publish because i must have an easy way to roll back a bad build.

fredrik--johansson:
Any news regarding this?

I would love to be able to automatically downgrade my users installed version.

My current strategy of reverting the faulty commit, build, and increase the version number is not ideal.

[...]

Issue #145, opened 2014-12-02
Fails to downgrade


kevinsawicki:
Was testing a downgrade from Atom 0.153 to 0.152 and it failed.

It appears to download 0.152 and then delete it, here is the log:

```
Program: Starting Squirrel Updater: --update https://atom.io/api/updates
Program: Starting update, downloading from https://atom.io/api/updates
CheckForUpdateImpl: Downloading RELEASES file from https://atom.io/api/updates
FileDownloader: Downloading url: 
CheckForUpdateImpl: hwhat, local version is greater than remote version
FileDownloader: Downloading file: 
ApplyReleasesImpl: Writing files to app directory: C:\Users\kevin\AppData\Local\atom\app-0.152.0
ApplyReleasesImpl: runPostInstallAndCleanup: starting fixPinnedExecutables
ApplyReleasesImpl: Processing shortcut ''
ApplyReleasesImpl: Does not match 'C:\Users\kevin\AppData\Local\atom\app-0.153.0', continuing to next directory
ApplyReleasesImpl: Processing shortcut 'C:\Program Files (x86)\Internet Explorer\iexplore.exe'
ApplyReleasesImpl: Does not match 'C:\Users\kevin\AppData\Local\atom\app-0.153.0', continuing to next directory
ApplyReleasesImpl: Squirrel Enabled Apps: [C:\Users\kevin\AppData\Local\atom\app-0.153.0\atom.exe]
ApplyReleasesImpl: cleanDeadVersions: for version 0.153.0
ApplyReleasesImpl: cleanDeadVersions: exclude folder app-0.153.0
ApplyReleasesImpl: cleanDeadVersions: exclude folder app-0.153.0
Utility: Starting to delete folder: C:\Users\kevin\AppData\Local\atom\app-0.152.0
```

douglasascosta:
How you fixed the downgrade? I have this problem too.

Thieum:
@kevinsawicki It looks more like an Atom issue than a squirrel issue? The downgrade process has to be supported by the application (no auto update mainly).

shiftkey:
I'm curious if this is still reproducible on a more recent version of Squirrel, because it looks like this part doesn't detect the version chosen to be installed:

```
ApplyReleasesImpl: cleanDeadVersions: for version 0.153.0
ApplyReleasesImpl: cleanDeadVersions: exclude folder app-0.153.0
[...]

PR #1484, opened 2019-05-07
docs: add rollback

gojanpaolo:
I think we can close the following issues:
#796 - nothing much here. just a duplicate
#964 - the posted solution here is a script to get the version for the next release. Nothing Squirrel-specific aside from the Releases folder.

That leaves us with two issues #524 and #145 to keep track of the rollback support. I'd prefer the #524 since it has attracted more attention from other users.

### DIFF
--- a/docs/using/update-process.md
+++ b/docs/using/update-process.md
@@ -15,8 +15,7 @@ The following steps are performed by the `UpdateManager` each time your app is e
 
 ## Rollback
 
-Currently, there is no built-in support for rolling back to a previous version. The suggested workaround is to release the previous stable version with a higher version number.
-Another workaround is to implement rollback on your own code base (see [sample](https://github.com/Squirrel/Squirrel.Windows/issues/524#issuecomment-450904627)).
+Currently, there is no built-in support for rolling back to a previous version.
 
 ## See Also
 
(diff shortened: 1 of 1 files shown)


docs/using/update-process.md
```diff
@@ -13,6 +13,11 @@ The following steps are performed by the `UpdateManager` each time your app is e
 4. **Update Shortcuts** - desktop and Windows Start Menu shortcuts are updated to point to the new MyApp version (via the `--processStart` command line parameter passed to `Update.exe`).
 5. **Previous Version Clean-up** - on the next startup of MyApp, all but current and immediately previous version of your app are deleted as part of clean up (e.g., after updating to app-1.0.5, app-1.0.4 will remain, but app-1.0.3 and before will be deleted - see [issue #589](https://github.com/Squirrel/Squirrel.Windows/issues/589)). 
 
+## Rollback
+
+Currently, there is no built-in support for rolling back to a previous version. The suggested workaround is to release the previous stable version with a higher version #.
+Another workaround is to implement rollback on your own code base (see [sample](https://github.com/Squirrel/Squirrel.Windows/issues/524#issuecomment-450904627)).
```
shiftkey: I worry about linking to this example that contains `TODO` statements and isn't necessarily recommended by the maintainers. Let's drop this comment for now and leave #524 open to track this discussion.

gojanpaolo: Removed rollback workaround

shiftkey: Let's hold off on any further doc changes in this PR, and try and consolidate these issues a bit more.

PR #1487, opened 2019-05-12
Modernize Squirrel build process

robmen:
Plus fixed a few whitespace and warnings

This is one step closer to automated releases.


Squirrel.sln
```diff
@@ -1,26 +1,20 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio Version 16
```
MeikTranel: Are you we should upgrade this already? The C++ projects in this solution are targeting the Windows 8.1 SDK, which does not come as part of VS2019. You have install manually from the SDK download pages. I'd at least expect a documentation entry in the contributing page.

shiftkey: > I'd at least expect a documentation entry in the contributing page.

Putting this on my list to dogfood as part of reviewing this PR locally

robmen: @MeikTranel does that matter? I don't have a machine with only VS2017 but my understanding was that VS2017 would open the .sln just fine. I just don't fight the .sln header.

Updating: the Windows SDK is next on my list but I need to get testing the native code more before being comfortable with that.

src/Setup/Setup.vcxproj
```diff
@@ -50,29 +36,22 @@
...
     <IncludePath>$(VC_IncludePath);$(WindowsSDK_IncludePath);$(ProjectDir)/wtl90</IncludePath>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <LinkIncremental>false</LinkIncremental>
     <IncludePath>$(VC_IncludePath);$(WindowsSDK_IncludePath);$(ProjectDir)/wtl90</IncludePath>
-    <OutDir>$(ProjectDir)bin\$(Configuration)\</OutDir>
-    <IntDir>$(ProjectDir)obj\$(Configuration)\</IntDir>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='CIBuild|Win32'">
+  <!-- <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='CIBuild|Win32'">
```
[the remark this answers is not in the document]
robmen: I'm not quite prepared to get rid of that code because it references the WinDDK. It will be resolved when I resolved the Windows SDK.

robmen:
> @robmen let me know what things I should focus on aside from seeing the build pass

Yeah, not much here to look at. Everything builds the same differently. Unit tests all pass which is my best gauge of correctness at this time.

Squirrel.sln
MeikTranel: There's really nothing of value in the VS2019 sln format for us. Theres filters but those dont make any sense with <20 projects and no separation of client server or something similar. VS2019 will load vs2017 solutions just fine... the other way around is questionable. Making breaking changes is pretty much the sole reason why they would want to version it.

let's go with @shiftkey on this. are you cool with me already basing my local dev branch on this or do you expect loads more changes before reintegrating back into master, @robmen ?

src/Directory.Build.props
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <EnableSourceLink Condition=" '$(NCrunch)' == '1' ">false</EnableSourceLink>
```
MeikTranel: A short explanation regarding that NCrunch condition would be nice. Like an XML comment above it to have some context - otherwise these build heavy repositories get really ethereal. Also why not true/false?

Squirrel.sln
robmen: @MeikTranel in my verification, nothing is broken. VS2017 works. VS2019 works. Can you share the errors you are seeing?

MeikTranel: Nono, theres nothing broken. It's just that VS2019 officially does not offer support for the Windows 8.1 SDK + something *might* get bungled by the format upgrade in the future.

robmen: 1. Windows 8.1 SDK is temporary. 
2. Something _always_ might get messed up (minor releases break me as much as major releases).

src/SyncReleases/SyncReleases.csproj
```diff
@@ -1,88 +1,34 @@
...
+    <PackageReference Include="ILRepack" Version="1.26.0" PrivateAssets="All" />
   </ItemGroup>
-  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <PostBuildEvent>cd "$(TargetDir)"
-"$(SolutionDir)packages\ILRepack.1.26.0\tools\ILRepack.exe" /internalize /out:$(TargetFileName).tmp $(TargetFileName) SharpCompress.dll Microsoft.Web.XmlTransform.dll Splat.dll Squirrel.dll Octokit.dll NuGet.Squirrel.dll
-del "$(TargetFileName)"
-ren "$(TargetFileName).tmp" "$(TargetFileName)"</PostBuildEvent>
+
+  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
```
MeikTranel: Why not use the regular Sdk-Element? https://docs.microsoft.com/en-us/visualstudio/msbuild/sdk-element-msbuild

robmen: Have you tried it?

src/Directory.Build.props
robmen: "SourceLink" is disabled when "NCrunch" is 1. Try running under NCrunch without the line.

src/SyncReleases/SyncReleases.csproj
MeikTranel: Works fine on my machine. Only thing thats freaking out is ILRepack requesting me to download .NET 2.0/3.5 weirdly enough. What sdks are you using (dotnet --list-sdks)? Maybe a global.json might make sense.

robmen: `ILRepack.exe` is old. Probably targets .NET Framework v2.0. I'm glad it tells you what you need to do.

PR #1492, opened 2019-05-17
Compare prerelease

Ben3eeE:
See https://github.com/atom/atom/issues/18417#issuecomment-492435520

Just opening a PR of what I have. It seems to work.

Let me know if there is any remaining work you need.


shiftkey:
> Let me know if there is any remaining work you need.

@Ben3eeE thanks for opening this up. I actually managed to stumble into this underlying problem while shipping an update to GitHub Desktop today (because it was a `-beta10`) and while the packaging issue is fixed the launcher shim issue is still unresolved.

Any pointers on how to test this issue has been fixed by the PR?

Ben3eeE:
> Any pointers on how to test this issue has been fixed by the PR?

To test the changes I just ran the StubExecutable project from Visual Studio after having manually created some expected folders. With a breakpoint in `wWinMain()` to check the result of these lines:

https://github.com/Squirrel/Squirrel.Windows/blob/bdfcd72033565431a04055dcbd525dd2de22f481/src/StubExecutable/StubExecutable.cpp#L96-L97

For example:
`app-1.39.0-beta9` and `app-1.39.0-beta10` gives the beta9 folder for these variables before this change and the beta10 folder after this change. I tested with both `beta` and `nightly`.

I don't know how I could test it out fully, for example to install Atom with these changes and see that it is working.

rafeca:
Hey @shiftkey / @Ben3eeE 👋

Do you folks know if there's more work needed in order to merge this? If so, anything I can help with?

This is the biggest blocker we have to ship Atom nightly releases, so would be cool to get this moving 😄

shiftkey:
> If so, anything I can help with?

@rafeca I'm mostly trying to make time to create a setup so I can verify the bug is fixed - hopefully later in the week is what I'm aiming for. If you can help out with that it'd be greatly appreciated!

rafeca:
Thanks @shiftkey for your response! I've tried building `Squirrel.Windows` from my Windows VM without much luck (I've just followed the instruction on https://github.com/Squirrel/Squirrel.Windows#building-squirrel but I get errors that seem to indicate that my Visual Studio version does not match the project's or something).

If I could somehow get the `Squirrel` binaries generated from this PR, I could use them for a custom version of `electron-winstaller` and test this PR on Atom directly.

Thieum:
@rafeca since the c++ project are still in MSVC2017, you could either install the community edition of Visual Studio 2019 or 2017  with the .NET Desktop payload, and the C++ payload, making sure this is checked in VS2017 install: 
image

or in VS2019 install:
image

The build tools could also be used instead of the full IDE with the same kind of options.

rafeca:
Thanks @Thieum ! I've just been able to build the binaries 😃 

I'll try to get a fork of `electron-winstaller` with these changes

rafeca:
@shiftkey I've updated the current nightly channel of Atom to use a custom version of `electron-winstaller` that embeds the binaries generated from this PR.

I've tested successfully that this PR fixes the issue of upgrading from `nightly-9` to `nightly-10` on Atom. There's detailed information about the test plan that I've followed in https://github.com/atom/atom/pull/19573

For now Atom is going to keep using that custom version of `electron-winstaller` but we'd like to switch back to the official library for the stable release which is going to happen in 2/3 weeks, do you know if by that time we could have the changes of this PR propagated to a new version of `electron-winstaller`?

I hope this helps 😄

PR #1504, opened 2019-06-19
Improved doc for build based on #1492

Thieum:
See #1492 
Gives more details for which option needs to be checked in the visual installer, since it's now standardized for the last 2 versions of Visual Studio.
I attempted to explain it in a way that won't require updates often. Not sure how to give more details about the version of VC++ without having something that would need frequent updates.


gojanpaolo:
@Thieum should we try this? https://devblogs.microsoft.com/setup/configure-visual-studio-across-your-organization-with-vsconfig/

Thieum:
@gojanpaolo that looks nice, never tried it in my projects. Will take a look, as I have personnal projects in mixed c++ / c# too.

robmen (approved):

PR #1510, opened 2019-07-05
Change output type to WinExe

Goxiaoy:
prevent console display


robmen:
Resolved with smaller diff in #1642. Thanks.

Issue #897, opened 2016-12-09
Squirrel dependency packages update


GraySerg:
Hi! Is it safe to update dependency packages of suirrel package to latest version?
For now VS suggest me to update 
DeltaCompressionDotNet from 1.0.0 to 1.1.0
Mono.Cecil from 0.9.6.1 to 0.9.6.4
Splat from 1.6.2 to 2.0.0

Thieum:
DeltaCompressionDotNet has been updated to 1.1.0

Thieum:
Request to update DeltaCompression to 2.0.0 - #1181 
Mono.Cecil tested ok with 0.9.6.4, not with 0.10.0 - #1337
multiple issues with old version of SharpCompress - #1253

stephenegriffin:
Porting over security note from #1253 so it's not lost: 

> There's an advisory on SharpCompress < 0.21 (adamhathcock/sharpcompress@80ceb1c)

peppy:
Don't forget #1362 (#1181 was closed in favour of this issue but this issue doesn't mention SharpCompress outside of follow-up comments).

shiftkey:
@peppy thanks for the extra context. I'll also link to this comment to remind me about this:

> The last time we tried [SharpCompress] it broke NuGet's garbage Zip parsing and broke a bunch of people

https://github.com/Squirrel/Squirrel.Windows/pull/1362#issuecomment-461094447

jeremy-bridges:
In order to support version 7.0.1 of Splat, Squirrel would need to require version 4.6.1 of the .NET Framework. Microsoft's guidance on this is below. Any feedback on making this move?

https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies

jmagly:
4.6.1 is 5 years old now, personally I would say go for it but I am just a user with only about 20 customers.

thecodetinker:
> Porting over security note from #1253 so it's not lost:
> 
> > There's an advisory on SharpCompress < 0.21 ([adamhathcock/sharpcompress@80ceb1c](https://github.com/adamhathcock/sharpcompress/commit/80ceb1c))

Hello - is there any progress on the SharpCompress update please? I am working to promote Squirrel for deploying applications within my company, but it's hard to make a case for it when it contains a known vulnerability. 

From reading the background I appreciate this isn't a two-minute fix, but it would be great to have an idea of the current status, if possible? Thanks!

[...]

PR #1511, opened 2019-07-05
Update SharpCompress to 0.23.0

Thieum:
based on #1362
see #897

### DIFF
--- a/src/Squirrel/BinaryPatchUtility.cs
+++ b/src/Squirrel/BinaryPatchUtility.cs
@@ -49,7 +49,7 @@ namespace Squirrel.Bsdiff
         /// <param name="output">A <see cref="Stream"/> to which the patch will be written.</param>
         public static void Create(byte[] oldData, byte[] newData, Stream output)
         {
-            // NB: If you diff a file big enough, we blow the stack. This doesn't 
+            // NB: If you diff a file big enough, we blow the stack. This doesn't
             // solve it, just buys us more space. The solution is to rewrite Split
             // using iteration instead of recursion, but that's Hard(tm).
             var ex = default(Exception);
@@ -113,7 +113,7 @@ namespace Squirrel.Bsdiff
             int eblen = 0;
 
             using (WrappingStream wrappingStream = new WrappingStream(output, Ownership.None))
-            using (var bz2Stream = new BZip2Stream(wrappingStream, CompressionMode.Compress))
+            using (var bz2Stream = new BZip2Stream(wrappingStream, CompressionMode.Compress, true))
             {
                 // compute the differences, writing ctrl as we go
                 int scan = 0;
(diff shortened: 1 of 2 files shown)


Keboo:
@shiftkey is there anything I can do that would help move this PR forward?

pcrosthwaite:
amy idea's on when this will be merged in? it would remove the current high severity vulnerability with sharpcompress

anaisbetts:
Once again, this PR will break Squirrel, we've tried it before

https://github.com/Squirrel/Squirrel.Windows/pull/1362#issuecomment-461094447

thecodetinker:
Hello,

I have been following this for about a year since the SharpCompress vulnerability was first flagged by our automated checker. I understand that Squirrel is now stuck between a rock and a hard place, but it would be good to gather some ideas on how the impasse might be resolved. One cohort of users is happy to live with the security vulnerability because the headache of manually updating is the greater problem. But for another group, the reverse is true.

I wonder if one way through would be to create Squirrel 2.0, which would allow a path for upgrading SharpCompress without breaking anyone on 1.8.x? Or maybe a parallel build that contains the newer library? I do understand that Squirrel got burned badly last time, but sticking with SharpCompress 17.1 for ever doesn't seem viable.

As always, much thanks to all contributors. I know this is a tricky one, and am grateful for all the hard work on the project!

anaisbetts:
The fundamental problem is that NuGet itself breaks on zip files generated by new SharpCompress - this isn't a "We don't want to ship breaking changes" issue. This is a "Squirrel doesn't work anymore" issue. 

Fundamentally, this security issue just does not apply to Squirrel - if you can control the zip file, you don't *need* a Zip exploit to gain privileges, you just....add an executable. This is not a security boundary we care about.

The viable options here are:

1. Move to some completely different Zip library that generates zip files that are acceptable to NuGet
1. Figure out why SharpCompress is breaking NuGet's zip parsing (difficult because it calls into a closed-source binary), and fix it / change options to make it work
1. Leave Squirrel as-is, and tell your security policy team to get a Xanax prescription :)

thecodetinker:
Oh I see, finally! That makes much more sense now. Are there items open for options 1 and 2? Meanwhile I'm off to the nearest chemist.

Thanks again.

eugbaranov:
For what it's worth, `nuget.exe` internally uses [`System.IO.Compression.ZipArchive`](https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.ziparchive) for building packages.
If I had to guess, the introduction of `SharpCompress` predates those APIs.

@anaisbetts, would you consider `ZipArchive` as a potential replacement of `SharpCompress`?

alexThoerner:
Hello, 

sorry to re-open this discussion and as a user of squirrel thank you for all the great work. Implementing auto-updates with squirrel is really easy and great fun and I really enjoy using it. 

However, I do not really feel comfortable using a vulnerable library in a product, although I understood the explanation why this does not apply to squirrel. So are there any plans to use another ZIP library as replacement for `SharpCompress` ?

Thank you!

Issue #1525, opened 2019-08-09
Invalid comparison of latestLocal/latestRemote ReleaseEntry instances


ensign-nemo:
**Squirrel version(s)**
Current

**Description**
Invalid (possibly redundant) comparison of Latest Local vs Latest Remote release entries.

**Expected behavior**
[Source code line](https://github.com/Squirrel/Squirrel.Windows/blob/bdfcd72033565431a04055dcbd525dd2de22f481/src/Squirrel/UpdateManager.CheckForUpdates.cs#L163)

If latest local current release matches the latest remote release, the function should quit early and log as such.

**Actual behavior**
When constructing the UpdateInfo instance declaring what releases to download, the code tests

`if (latestLocal == latestRemote)`

This is comparing _instances_, not values as there is no == operator or .Equals override on ReleaseEntry. Thus two instances which actually do point to the same release do not satisfy the intent of the check.

PR #1526, opened 2019-08-09
#1525 Update UpdateManager.CheckForUpdates.cs

ensign-nemo:
Test for Version equality, rather than instance equality.


MungeParty:
Comparing SHA1 rather than Version would be more accurate, since it's possible to stomp a version.

PR #1554, opened 2019-10-11
UpdateManager.RestartApp fails when EntryAssembly is a dll

### LINKED ISSUE #1553: UpdateManager.RestartApp fails when EntryAssembly is a dll
We are using Squirrel.Windows in a WPF .NET Core 3 project.
It's working fine. We just had to add the Squirrel.dll and NuGet.Squirrel.dll manually to the project. The other nuget dependencies were added as normal nuget packages.

But when calling UpdateManager.Restart app it fails due to that Assembly.GetEntryAssembly() returns the dll for our WPF application, not our exe.

"Process.GetCurrentProcess().MainModule.FileName" works in our case and it is my understanding that it will always give the exe file of the main process, which is exactly what is wanted. See my pull request for this fix.

jeremia:
This resolves #1553


GeertvanHorrik:
Question: why not check if extension is .dll? If so, call Path.ChangeExtension to exe. That's how I solved the restarting of my .NET Core apps (that now are .dll entry assemblies). But if you have a good reason, I might have to change my method everywhere in my code :-)

jeremia:
Why check the extension and have an if/else instead of using a method that will work for both cases?
Keep It Simple and Stupid. 🙂

jeremia:
Since this has been open for 5 years + without getting merged and it's not building when I update the branch I will l close this.

PR #1581, opened 2020-01-24
DOC: GitHubUpdateManager

lightlike:
Update example for GitHubUpdateManager

Move `Task.Result` into `using` so the `UpdateManager` is disposed instead of the `Task`
(throws error otherwise)

For working example see [here](https://github.com/Drachenhorn-Team/Drachenhorn/blob/master/Drachenhorn.Desktop/UserSettings/SquirrelManager.cs)


lightlike:
Duplicate of #1396

PR #1552, opened 2019-10-11
Bug fix: Allowing spaces in exe file for UpdateManager.RestartApp.

jeremia:
This fix resolves #1257

PR #1592, opened 2020-02-24
On RestartApp enclose exeToStart with quotes

bertstomphorst:
On RestartApp enclose exeToStart with double-quotes when needed (exe-name contains spaces, and it's not contained within double-quotes by the caller)

I encountered a problem that my app didn't restart. This could be fixed by this change, and may help others also.


anaisbetts:
I went with the simpler fix in #1552, thanks for the PR though!

bertstomphorst:
I'm not sure if the other fix (#1552) doesn't break applications where the exeToStart is provided with double quotes around it?

PR #1623, opened 2020-05-15
Fixups for Squirrel.Windows 2.0

anaisbetts:
Creating a PR to prepare for Squirrel.Windows 2.0.0 which mostly merges #1578 but also fixes #1616. Since we're moving namespaces around, this is a breaking change but ideally this won't result in any _actual_ work for people.

## TODO:

- [x] Get the heckin' thing to build
- [x] Expose logging types in a better way since it's a public'ish API
- [x] Make sure the C# API changes don't break clients too horribly
- [x] Make sure that electron-winstaller can still build installers
- [x] Make sure that an update from a Squirrel.Windows 1.x installation doesn't explode
- [x] Verify that none of the SDK changes bork Windows 7 installers
- [x] Debug the CMD window showing up

/cc @shiftkey @robmen @jaybeavers @dennisameling

PR #1642, opened 2020-08-08
Hide console (CMD) when running Update.exe

dennisameling:
Problematic behavior described here: https://github.com/Squirrel/Squirrel.Windows/pull/1623#issuecomment-632362767

This PR resolves that problem 🚀

### DIFF
--- a/src/Update/Update.csproj
+++ b/src/Update/Update.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <TargetFramework>net45</TargetFramework>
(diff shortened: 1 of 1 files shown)


dennisameling:
CI is failing due to Splat dependency. https://github.com/Squirrel/Squirrel.Windows/pull/1623 will fix that, so this PR should be safe to merge either before or after #1623 is merged.

robmen:
@dennisameling  Thanks for finding that.

PR #1655, opened 2020-08-28
Fix #1654

### LINKED ISSUE #1654: UpdateApp() method always return null after upgrade to 2.0.0
**Squirrel version(s)**
v2.0.0

**Description**
mgr.UpdateApp() method always return null, even the ReleasesToApply list is not empty.

I found an exception was thrown here: 
https://github.com/Squirrel/Squirrel.Windows/blob/develop/src/Squirrel/IUpdateManager.cs#L171

```text
Could not load file or assembly 'Mono.Cecil, Version=0.11.2.0, Culture=neutral, PublicKeyToken=50cebf1cceb9d05e' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```

**Expected behavior**
Return a valid ReleaseEntry

**Actual behavior**
Always return null

Seems that the package Mono.Cecil referenced by [Squirrel.Windows](https://www.nuget.org/packages/squirrel.windows/2.0.0) need to upgrade.

yunfandev:
Fix #1654

### DIFF
--- a/src/Squirrel.nuspec
+++ b/src/Squirrel.nuspec
@@ -10,7 +10,7 @@
 
     <dependencies>
       <dependency id="DeltaCompressionDotNet" version="[1.1,2.0)" />
-      <dependency id="Mono.Cecil" version="0.9.6.1" />
+      <dependency id="Mono.Cecil" version="0.11.2" />
       <dependency id="SharpCompress" version="[0.17.1]" />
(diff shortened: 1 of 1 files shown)


anaisbetts:
Must have been a merge mistake, thanks for the fixup

PR #1656, opened 2020-08-31
Add percentage calculation

### LINKED ISSUE #1652: Improve progress reporting
We have fairly large apps. What we see whenever we run an update:

1. Instant download (since delta package is normally ~ 2-4 mb)
2. Squirrel being "stuck" at 66%

Then completing very fast again.

I've been digging into this, and think it's caused by the full packages from delta creation:

https://github.com/Squirrel/Squirrel.Windows/blob/813dd91a5ec179cb4e2db9b692fd34296c5db599/src/Squirrel/UpdateManager.ApplyReleases.cs#L35

I think the apply releases progress can be split into the following:

```
00 => start
00 => 40 (create full packages from deltas)
40 => 80 (install pacakge to dir)
80 => 85 (update local releases file)
85 => 90 (self update)
90 => 95 (post install)
95 => 96 (remove dead entries)
96 => 97 (unshim ourselves)
96 => 98 (clean dead versions)
100
```

I think this will more reliably reflect the actually installation status and give users a better sense of progress.

**I am happy to implement this (non-breaking) change as a PR if you are open to this.**

GeertvanHorrik:
Fixes #1652

### DIFF
--- a/src/Squirrel/ApplyReleasesProgress.cs
+++ b/src/Squirrel/ApplyReleasesProgress.cs
@@ -1,0 +1,46 @@
+﻿namespace Squirrel
+{
+    using System;
+
+    internal class ApplyReleasesProgress : Progress<int>
+    {
+        private readonly int _releasesToApply;
+        private int _appliedReleases;
+        private int _currentReleaseProgress;
+
+        public ApplyReleasesProgress(int releasesToApply, Action<int> handler)
+            : base(handler)
+        {
+            _releasesToApply = releasesToApply;
+        }
+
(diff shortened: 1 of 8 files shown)


GeertvanHorrik:
Integration test progress example (shows correct calculation of applying releases from 40 => 80):

image

GeertvanHorrik:
And progress with create delta packages:

image

GeertvanHorrik:
Done, requesting review.

anaisbetts:
👍

Issue #1586, opened 2020-02-07
Unistall leaves .dead file in the %localappdata% that prevents new installations


SuppaDev:
**Squirrel version(s)**
Using AutoSquirrelUI 1.1.27

**Description**
Installation, and Update work great, however when you choose to uninstall the application it takes you to the control panel. Uninstalling works as intended but leaves the %localappdata%\App folder with Update.exe and .DEAD files. The presence of this .DEAD file will cause the installer of the same/or new versions to fail

**Steps to recreate**
1 - Create c# 8.0 Application, utilizing PackageReference and SDK style csproj
2 - Package using the SquirrelAutoUI
3 - Install version using Setup.exe
4 - Uninstall version using the control panel
5 - Attempt to install using the same Setup.exe
6 - Fails out 
[SquirrelSetup.log](https://github.com/Squirrel/Squirrel.Windows/files/4171039/SquirrelSetup.log)

**Expected behavior**
Uninstaller would either remove the directory entirely, OR installer would validate .dead file and handle exceptions installing

**Actual behavior**
Installer exceptions out a second run of the installer works as intended

**Additional information**
File added above

PR #1667, opened 2020-10-08
Fix issue where exe files were not removed during uninstall.

### LINKED ISSUE #1641: GetAllSquirrelAwareApps locks all exe files, therefore the uninstall process can't delete these exe files
**Squirrel version(s)**
1.9.2

**Description**
uninstall process locks the exe files in the app dir, so it can't remove them, which ends up in a folder which isn't empty

**Steps to recreate**
uninstall a squirrel installation which contains multiple exe files

**Expected behavior**
exe files should be removed from the app folder

**Actual behavior**
exe files aren't being removed from the app folder

**Additional information**
I've isolated the issue to this line, in GetAllSquirrelAwareApps method:
            return di.EnumerateFiles()
                .Where(x => x.Name.EndsWith(".exe", StringComparison.OrdinalIgnoreCase))
                .Select(x => x.FullName)
                .Where(x => (GetPESquirrelAwareVersion(x) ?? -1) >= minimumVersion)
                .ToList();
Something in this line must be locking the files

johan-lindqvist:
This error occurred due to a disposable resource not being disposed.

Tested with our application and the change resolves the issue of exe-files not being removed since version 2.0.0.

Fixes #1641 

Sidenote: there's still two files left after an uninstall, a ".dead"-file, and Update.exe as described in #1586, but if I understand correctly that is expected behavior?

### DIFF
--- a/src/Squirrel/SquirrelAwareExecutableDetector.cs
+++ b/src/Squirrel/SquirrelAwareExecutableDetector.cs
@@ -36,25 +36,26 @@ namespace Squirrel
         static int? GetAssemblySquirrelAwareVersion(string executable)
         {
             try {
-                var assembly = AssemblyDefinition.ReadAssembly(executable);
-                if (!assembly.HasCustomAttributes) return null;
+                using (var assembly = AssemblyDefinition.ReadAssembly(executable)) {
+                    if (!assembly.HasCustomAttributes) return null;
 
-                var attrs = assembly.CustomAttributes;
-                var attribute = attrs.FirstOrDefault(x => {
-                    if (x.AttributeType.FullName != typeof(AssemblyMetadataAttribute).FullName) return false;
-                    if (x.ConstructorArguments.Count != 2) return false;
-                    return x.ConstructorArguments[0].Value.ToString() == "SquirrelAwareVersion";
-                });
+                    var attrs = assembly.CustomAttributes;
+                    var attribute = attrs.FirstOrDefault(x => {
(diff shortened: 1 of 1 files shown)


GeertvanHorrik:
Looks good to me!

jltjohanlindqvist:
@anaisbetts any chance of getting this reviewed? We currently use a forked version of this but it would be nice to have it in the main package.

anaisbetts:
Please don't reformat the code according to some other formatting style

src/Squirrel/SquirrelAwareExecutableDetector.cs
```diff
@@ -29,32 +29,34 @@ public static List<string> GetAllSquirrelAwareApps(string directory, int minimum
...
                 GetAssemblySquirrelAwareVersion(fullname) ?? GetVersionBlockSquirrelAwareValue(fullname));
         }
 
         static int? GetAssemblySquirrelAwareVersion(string executable)
         {
             try {
-                var assembly = AssemblyDefinition.ReadAssembly(executable);
-                if (!assembly.HasCustomAttributes) return null;
+                using (var assembly = AssemblyDefinition.ReadAssembly(executable))
+                {
```
anaisbetts, line 40: :arrow_up:, the bracket belongs on the same line

jltjohanlindqvist: Updated to what I think you meant.

Issue #1050, opened 2017-06-07
How can I use UpdateManager.CreateShortcutsForExecutable with onlyUpdate = true


shimika:
I want to approach onlyUpdate to true by call update.exe but it seems no method.

There is only

```
mgr.CreateShortcutsForExecutable(exeName, locations ?? defaultLocations, false, processStartArgs, icon);
```
maybe

Thieum:
@shimika  There is a comment in the code explaining the behavior of this flag when used with CreateShortcutsForExecutable: https://github.com/Squirrel/Squirrel.Windows/blob/de5baa93a134ac24fd5bf9172d3e457ac27e0b7a/src/Squirrel/UpdateManager.ApplyReleases.cs#L214 

```
                    // NB: If we've already installed the app, but the shortcut
                    // is no longer there, we have to assume that the user didn't
                    // want it there and explicitly deleted it, so we shouldn't
                    // annoy them by recreating it.
```
Does it clarify the intent of this flag?

vidia:
I have a similar desire as the OP here. 

I'm trying to update the shortcuts for a project installed by squirrel, and I'd like to leverage this option for `CreateShortcutsForExecutable` to only update the shortcuts (so we don't end up recreating shortcuts after the user has deleted them). Currently we execute squirrel with `--createShortcut=` to generate the shortcuts in the start menu and desktop. Unfortunately though without making additional changes, this means we end up recreating the shortcuts even if a user deletes them. 

Would it be reasonable to expose this option under the `--createShortcut=` flag somehow? Possible as an additional flag `--only-update` following it? I'd like to avoid recreating the same logic in my own update code if I can build on what already exists in squirrel.

Issue #831, opened 2016-09-18
Squirrel.exe does not support command line option to avoid recreating deleted shortcut


aekeus:
The third parameter is a static `false` value in the call to update / create a shortcut `CreateShortcutsForExecutable`. It would be helpful if a command line option was available to pass instead.

https://github.com/Squirrel/Squirrel.Windows/blob/ee32fb469fc1ace0af6aef7edc389e27b8c063bb/src/Update/Program.cs#L464

vidia:
#1050 is related to this as well.

PR #1674, opened 2020-10-22
Added CLI flag to onlyUpdate shortcuts

vidia:
By default, running with `--createShortcut` will create shortcuts in each specified location. Running it again after a user has explicitly deleted a shortcut will result in the shortcut being re-created. If the user has deleted one of these shortcuts, Squirrel should be able to optionally not recreate them in order to respect the user's decision to delete. Because the calculation of shortcut paths and details is handled within Squirrel, this should be handled by Squirrel itself rather than by Squirrel apps. 

`CreateShortcutsForExecutable` has an option to "onlyUpdate" shortcuts (i.e. skipping if the shortcut does not already exist), this change exposes this option to the CLI as a flag.

This is intended to resurrect #1368 which has gone out of date. 

This impacts the following issues: #1050, #831 and potentially enables a solution for multiple other shortcut related issues (see references in linked issues).

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -122,7 +122,7 @@ namespace Squirrel.Update
                     UpdateSelf().Wait();
                     break;
                 case UpdateAction.Shortcut:
-                    Shortcut(opt.target, opt.shortcutArgs, opt.processStartArgs, opt.setupIcon);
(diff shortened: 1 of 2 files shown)


dylanclark:
Would also like to see this exposed, thanks for making a PR for this!

anaisbetts:
👍 , thanks!

PR #1682, opened 2020-11-25
Modernise

roederja2:
* Target .net framework 4.6.2
* Update dependencies
* Build update.exe as Exe rather than WindowsExe so we actually get console output

I saw the problems you had earlier with updating SharpCompress. Maybe the move to .net framework 4.6.2 fixes this. I had no issues updating my installation with this version


mungojam:
I haven't looked at the innards much. Why not go up to 4.8 or even .net 5?

PR #1721, opened 2021-05-09
Add flag to bypass UAC check for wine

### LINKED ISSUE #1720: feature request: Add flag to allow for bypassing UAC / admin check
Motivation: In Wine, all application appear to run as admin, which prevents programs using Squirrel from being installed. For example, https://github.com/goatcorp/FFXIVQuickLauncher/issues/303

This comes from the logic here: https://github.com/Squirrel/Squirrel.Windows/blob/76c87af6f389bea69576c738ef46574b2945b0ba/src/Setup/winmain.cpp#L77-L78, implemented in https://github.com/Squirrel/Squirrel.Windows/pull/254.

I'm guessing people have caused permission issues before by doing weird stuff with UAC, so I'd propose naming it something scary or very specific, like `--dangerouslyBypassUAC` or `--wineUserFix`. Hopefully hiding it behind a flag will prevent people from tripping on it.

A note on wine: Yeah, this should probably be fixed in Wine rather than here, but their workaround (runas.exe) has been disabled since wine-staging 5.16+ so right now nothing can be done. Relevant Wine bug report: https://bugs.winehq.org/show_bug.cgi?id=40613

oslerw:
Allow for bypassing the UAC check to fix a Wine compatibility issues. Fixes #1720

### DIFF
--- a/src/Setup/winmain.cpp
+++ b/src/Setup/winmain.cpp
@@ -73,6 +73,7 @@ int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
 	bool isQuiet = (cmdLine.Find(L"-s") >= 0);
 	bool weAreUACElevated = CUpdateRunner::AreWeUACElevated() == S_OK;
 	bool attemptingToRerun = (cmdLine.Find(L"--rerunningWithoutUAC") >= 0);
+	bool skipUACCheck = (cmdLine.Find(L"--wineUserFix") >= 0);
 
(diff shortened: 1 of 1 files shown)


anaisbetts:
While I definitely think that the Wine bug is legit justification to do *something*, by the time this is fixed in Squirrel and apps pick it up, Wine will have already fixed this. Wine is a faster place to make this kind of workaround

PR #1732, opened 2021-08-08
Add support for language neutral VersionInfo in target .exes

mungojam:
A partial fix for #1644, by adding support for language neutral VersionInfo entry, rather than just en-US.

Consumers who wish to use this will need to add the SquirrelAware attribute to the .net core exe as a post build step using a tool like rcedit (already present in the squirrel library).

The fully automated solution for #1644 is still under discussion in #1692.

@ComtelJeremy


mungojam:
@anaisbetts, here is the first part of the original .net core PR (#1692), simply adding support for language neutral exes which then covers the .net core launcher exes.

PR #1743, opened 2021-09-28
Mention Google Omaha in README

mherrmann:


robmen:
Thanks but doesn't seem worth the footnote. If you want to update your PR comment with a persuasive argument on how the footnote helps, could re-entertain.

PR #1744, opened 2021-09-29
Fixing dependabot security warnings

danworley:
Dependabot compliance is required to maintain security SLAs with various third-parties. These are the changes required to get Squirrel.Windows into compliance.
- Updated projects to .NET Framework 4.7.2 for .NET Standard compatibility
- Updated SharpCompress to 0.29.0 for Dependabot warns (Requires .NET Standard).
- Included netstandard reference.
- Implemented default value of false for BZip2Stream to maintain existing functionality with SharpCompress.


anaisbetts:
This has been discussed a bunch, we cannot update SharpCompress because it breaks updates

danworley:
Oh, I wasn't aware of that. How does it break them? In my (limited) testing it worked ok. Is there something I'm missing? I was getting ready to push out a release with this fix so any info would be appreciated! Thanks!

anaisbetts:
> Oh, I wasn't aware of that. How does it break them?

As I recall, SharpCompress generates Zip archives that NuGet's Zip decompressor inexplicably can't read - you send down the update and it fails to install, meaning all your users are stranded

PR #1746, opened 2021-10-03
Update VC toolset to 14.2 (VS 2019)

AArnott:
This is what modern Visual Studio installations tend to install components for, making this repo more approachable.


AArnott:
This is blocking #1748 which is the last thing to warning-free use of squirrel by .NET Core apps.

robmen (approved):

PR #1747, opened 2021-10-03
Update and consolidate NB.GV refs

AArnott:
An old NB.GV was referenced everywhere via Directory.Build.props file. Newer ones were updated in specific projects, although I don't know why the root level one was not updated instead.
This change removes all the Update elements in favor of updating the one root-level one.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -17,7 +17,7 @@
 
   <ItemGroup Condition=" '$(MSBuildProjectExtension)'=='.csproj' ">
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All" />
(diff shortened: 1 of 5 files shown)


AArnott:
One question for the maintainers: I see the version of the nuget package built in this repo has no bearing on the versions stamped into the DLLs. The version of the package is hard-coded in the .nuspec file. I don't see any pipeline step, script, or document about how this package is to be constructed, which seems to be to "build first, then run `nuget pack src`". I could introduce an msbuild project (or reuse an existing one) that can self-pack after building all dependencies. It would also be self-versioned. Or I could write a script that would use `nbgv get-version' and then invokes `nuget pack src -Version %version%` based on that. 

Thoughts?

robmen:
I set that up originally and Ana decided to stick to manually managing the version in Squirrel 2. In Squirrel 3, we can re-evaluate the versioning scheme again.

robmen (approved):

PR #1748, opened 2021-10-03
Drop DeltaCompressionDotNet dependency

AArnott:
This may be reverted after https://github.com/taspeotis/DeltaCompressionDotNet/issues/17 is fixed.
Closes #1716

PR #1745, opened 2021-10-03
Add netstandard2.0 target to squirrel

AArnott:
Fixes #1649

PR #1751, opened 2021-10-04
Really target netstandard2.0, and drop Delta dependencies

AArnott:
- Really drop the DeltaCompressionDotNet dependency
  In #1748 we removed it from code and the build. But I forgot to remove it from the .nuspec file till now.
- Although in #1745 we built for netstandard2.0, I forgot to update the .nuspec to indicate this.

### DIFF
--- a/src/Squirrel.nuspec
+++ b/src/Squirrel.nuspec
@@ -9,9 +9,14 @@
     <iconUrl>https://raw.githubusercontent.com/Squirrel/Squirrel.Windows/master/docs/artwork/Squirrel-Logo-Square.png</iconUrl>
 
     <dependencies>
-      <dependency id="DeltaCompressionDotNet" version="2.0.0" />
-      <dependency id="Mono.Cecil" version="0.11.2" />
-      <dependency id="SharpCompress" version="[0.17.1]" />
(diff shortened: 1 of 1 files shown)


src/Squirrel.nuspec
```diff
@@ -9,9 +9,14 @@
     <iconUrl>https://raw.githubusercontent.com/Squirrel/Squirrel.Windows/master/docs/artwork/Squirrel-Logo-Square.png</iconUrl>
 
     <dependencies>
-      <dependency id="DeltaCompressionDotNet" version="2.0.0" />
-      <dependency id="Mono.Cecil" version="0.11.2" />
-      <dependency id="SharpCompress" version="[0.17.1]" />
+      <group targetFramework=".NETFramework4.5">
```
robmen, line 12: Why group these separately? They are the same, no?

AArnott: It isn't strictly required I think, but `nuget pack` warns when you don't have a distinct dependencies group for every TFM. Once I defined a `lib\netstandard2.0` folder in you nuspec, a new pack warning appeared until I did this.

theresia-tobii:
When will a package be released with this fix? Latest one (2.1.0 is from dec 2020)

PR #1752, opened 2021-10-04
Fix `CreateShortcutForThisExe` for .NET Core apps

AArnott:
The entrypoint assembly is reported as being a .dll in the .NET Core case. A shortcut to the dll is *not* what the developer expects and it doesn't work for the user. Instead, look for a nearby .exe with the same file name and use that.

### DIFF
--- a/src/Squirrel/IUpdateManager.cs
+++ b/src/Squirrel/IUpdateManager.cs
@@ -183,8 +183,17 @@ namespace Squirrel
 
         public static void CreateShortcutForThisExe(this IUpdateManager This)
         {
-            This.CreateShortcutsForExecutable(Path.GetFileName(
-                Assembly.GetEntryAssembly().Location),
+            string entrypoint = Assembly.GetEntryAssembly().Location;
+            if (String.Equals(Path.GetExtension(entrypoint), ".dll", StringComparison.OrdinalIgnoreCase)) {
+                // This happens in .NET Core apps. A shortcut to a .dll doesn't work, so replace with the .exe.
+                string candidateExe = Path.Combine(Path.GetDirectoryName(entrypoint), Path.GetFileNameWithoutExtension(entrypoint)) + ".exe";
+                if (File.Exists(candidateExe)) {
+                    entrypoint = candidateExe;
+                }
+            }
+
+            This.CreateShortcutsForExecutable(
+                Path.GetFileName(entrypoint),
                 ShortcutLocation.Desktop | ShortcutLocation.StartMenu, 
                 Environment.CommandLine.Contains("squirrel-install") == false,
                 null, null);


anaisbetts:
So, someone else proposed this solution as well and I'm not super excited about it, because lots of other projects also have DLLs with the same name as EXEs. If you can come up with a better way of detecting the .NET Core EXE stub (perhaps by peeking in the PE header bytes) that'd be Better than just seeing the DLL with the same name

AArnott:
Thanks for looking, @anaisbetts. But I think you may be confusing this fix with #1692, which fixes a different problem and jumps from .exe to .dll. I think your concern *there* has some validity because you're jumping from an .exe which may itself be legit to a dll. 
But here in this PR, we're going from an *invalid* .dll selection to a *valid* .exe selection. I explain in the PR description and code comments just why this is a safe move: selecting a .dll for a shortcut is broken, so this change can't make anything worse. And unless you know of another situation in which `GetEntryAssembly` returns a .dll instead of an .exe then I think we can call this safe. There won't be anything in the .dll to point to the .exe because the .exe is built long after the .dll is during the build, so our solution has to be pattern based. I also check that the .exe exists for added caution.

anaisbetts:
@AArnott That works for me, fix the style to match and I'm sold (`string.Equals` => `String.Equals`, put `{` on same line as the `if`)

AArnott:
Done

src/Squirrel/IUpdateManager.cs
```diff
@@ -183,8 +183,18 @@ public static async Task<ReleaseEntry> UpdateApp(this IUpdateManager This, Actio
...
         public static void CreateShortcutForThisExe(this IUpdateManager This)
         {
-            This.CreateShortcutsForExecutable(Path.GetFileName(
-                Assembly.GetEntryAssembly().Location),
+            string entrypoint = Assembly.GetEntryAssembly().Location;
+            if (String.Equals(Path.GetExtension(entrypoint), ".dll", StringComparison.OrdinalIgnoreCase)) {
+                // This happens in .NET Core apps. A shortcut to a .dll doesn't work, so replace with the .exe.
+                string candidateExe = Path.Combine(Path.GetDirectoryName(entrypoint), Path.GetFileNameWithoutExtension(entrypoint)) + ".exe";
+                if (File.Exists(candidateExe))
+                {
```
anaisbetts, line 191: ⬆️

AArnott: 🤦‍♂️
An .editorconfig file to help conform to this would be *great*.
Fixed.

anaisbetts: For sure, I should add that

Thieum: @anaisbetts see https://github.com/Squirrel/Squirrel.Windows/pull/1482 for an attempt.

PR #1799, opened 2022-04-13
Show signtool output on error

mjerde:
The signtool output is captured but only displayed in the event of success. This makes it hard to identify the causes of signtool problems, especially since signtool is operating on files in a temp location which are cleaned up after failure. Adding output of the signtool output on error helped me identify the cause of a problem I had when trying to releasify.

Thank you, Squirrel team!

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -596,6 +596,8 @@ namespace Squirrel.Update
                 String.Format("sign {0} \"{1}\"", signingOpts, exePath), CancellationToken.None);
 
             if (processResult.Item1 != 0) {
+                Console.WriteLine("signtool failed. output:");
+                Console.WriteLine(processResult.Item2);
                 var optsWithPasswordHidden = new Regex(@"/p\s+\w+").Replace(signingOpts, "/p ********");
                 var msg = String.Format("Failed to sign, command invoked was: '{0} sign {1} {2}'",
                     exe, optsWithPasswordHidden, exePath);


caesay:
This has been fixed (along with several other signing related issues) [in my fork here](https://github.com/clowd/Clowd.Squirrel).

src/Update/Program.cs
```diff
@@ -596,6 +596,8 @@ static async Task signPEFile(string exePath, string signingOpts)
                 String.Format("sign {0} \"{1}\"", signingOpts, exePath), CancellationToken.None);
 
             if (processResult.Item1 != 0) {
+                Console.WriteLine("signtool failed. output:");
+                Console.WriteLine(processResult.Item2);
```
robmen, line 600: Instead of writing to the Console how about adding the output to the exception message instead?

robmen (changes requested):

src/Update/Program.cs
anaisbetts: 👍

anaisbetts: My only question is, are we sure that signtool output never leaks secrets?

robmen: All good. It does not.

mjerde: Since line 602 is already identical to this line, called in the non-error case, I would say this (writing to Console) is consistent with the program as it is already.

My only concern with adding it to the exception is that it would lend itself to making the error difficult to read as the exception text gets larger. Being some years since I was a daily C# dev, I don't presently have the expertise to identify the full range of possible effects for a large exceptions text but feel comfortable assessing that captured STDOUT will retain readability when written to Console.

RADOVSHOP: - - [ ] @~~~~

mjerde:
Closed due to lack of traction.

PR #1800, opened 2022-04-13
Fix 'debug mode' pathToWixTools

mjerde:
The "debug mode" search for `candle.exe` in `pathToWixTools` causes the method to return the full path of the `candle.exe` while the other search returns the path of the folder containing `candle.exe`.

Prior to this PR it results in attempts to use `vendor\wix\candle.exe\template.wxs` rather than `vendor\wix\template.wxs`.

Thank you, Squirrel team! (This is the last of the PRs I have planned. They seemed unrelated enough to be inappropriate to pack together into a single PR. Thanks!)

### DIFF
--- a/src/Update/Program.cs
+++ b/src/Update/Program.cs
@@ -744,8 +744,8 @@ namespace Squirrel.Update
             }
 
             // Debug Mode (i.e. in vendor)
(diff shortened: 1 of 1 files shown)


robmen:
I very much appreciate the separate PRs.

robmen (approved):

PR #1823, opened 2022-07-13
TLS 1.3 support

sebastian-jtl:
Hard setting the security protocols in the ServicePointManager disallows the usage of newer security protocols for the update source.
Enabling TLS 1.3 support in the app using Squirrel should now work as intended and not be overwritten whenever a web client is created.

### DIFF
--- a/src/Squirrel/Utility.cs
+++ b/src/Squirrel/Utility.cs
@@ -104,8 +104,15 @@ namespace Squirrel
 
         public static WebClient CreateWebClient()
         {
-            // WHY DOESNT IT JUST DO THISSSSSSSS
-            System.Net.ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls;
+            // enable TLS support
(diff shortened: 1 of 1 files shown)


JTLBaumann (approved):

caesay:
TLS 1.0 and 1.1 are officially deprecated, and it's Microsoft's advice to disable them at the OS level. Seems like it may be beneficial to leave them alone (eg. `|=` Tls 1.3 only & remove Ssl)

caesay:
Well fundamentally I agree with you, Squirrel probably shouldn't be touching ServicePointManager. Where I disagree with you is this being "safe". 

If the user/developer has already removed TLS1.0/1.1 in the startup of their app, using the following code for example:

```cs
ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
```

As soon as Squirrel runs, it will add back TLS1.0 and 1.1 which are unsafe protocols. If Squirrel is going to touch the ServicePointManager at all, the only thing we should be doing is adding TLS1.2 / TLS1.3 and nothing else.

sebastian-jtl:
@caesay I meant "safe" as in the way that it does not change the current behavior of what security protocols will be available by default after the patch, not as in security. I would fully support completely removing the call.

PR #1827, opened 2022-08-21
Add title, body and draft release flag to CLI.

sirdoombox:
- Added title, body and draft release flag to the github-up command
- Included release notes in the body text after the body - allows you to include some fixed information in every build without adding it to every release file (not sure if I did this 100% correctly, though it seems right to me)
- Switched release check to check tags instead of title, much more reliable source and allows you to freely rename the title of the release as you see fit.


sirdoombox:
Sorry, entirely wrong repo.

PR #1830, opened 2022-09-16
fixes #974

### LINKED ISSUE #974: SignTool password masking fails with special characters
The password masking implemented in #861 fails if the password contains a special character. 

For example a password of 1@234567 will show in the logs as ********@234567

H-Ashrafi:


caesay:
This merged regex has unescaped forward slashes and a non-fixed length lookbehind. I didn't test if this works with dotnet but it probably won't with some regex implementations.

PR #1862, opened 2023-07-28
Fix app can not restart after an unexpected exit during the apply releases

DamonYu6:
Problem Description: During the use of Squirrel's automatic upgrade feature, if the app is quit or unexpectedly exits while installing the new version, it may result in unable to restart the app. Upon reviewing the source code, it was found that the stubExe would find the latest version and create the path to launch it. However, due to missing files in the latest app-xxx directory, the app fails to start.

Solution: To address this issue, during the application of releases, the new version is first extracted to a temp-app-xxxx directory. After the extraction is complete, it is then renamed to the final app-xxxx directory. This approach ensures that even if the app unexpectedly exits during the installation process and leaves the temp directory, it will not affect the next app launch.

### DIFF
--- a/src/Squirrel/UpdateManager.ApplyReleases.cs
+++ b/src/Squirrel/UpdateManager.ApplyReleases.cs
@@ -299,30 +299,30 @@ namespace Squirrel
             {
                 return Task.Run(async () => {
                     var target = getDirectoryForRelease(release.Version);
-                    var tempTarget = getTempDirectoryForRelease(release.Version);
 
                     // NB: This might happen if we got killed partially through applying the release
-                    if (tempTarget.Exists) {
-                        this.Log().Warn("Found partially applied temp release folder, killing it: " + tempTarget.FullName);
-                        await Utility.DeleteDirectory(tempTarget.FullName);
-                    }
-
-                    tempTarget.Create();
-
-                    this.Log().Info("Writing files to temp app directory: {0}", tempTarget.FullName);
-                    await ReleasePackage.ExtractZipForInstall(
-                        Path.Combine(updateInfo.PackageDirectory, release.Filename),
-                        tempTarget.FullName,
-                        rootAppDirectory,
-                        progressCallback);
-                    
                     if (target.Exists) {
                         this.Log().Warn("Found partially applied release folder, killing it: " + target.FullName);
                         await Utility.DeleteDirectory(target.FullName);
                     }
 
-                    this.Log().Info("Rename directory {0} to {1}", tempTarget.FullName, target.FullName);
-                    tempTarget.MoveTo(target.FullName);
+                    target.Create();
+
+                    // Create the .not-finished file before extraction is started
+                    var notFinishedFilePath = Path.Combine(target.FullName, ".not-finished");
+                    File.WriteAllText(notFinishedFilePath, "");
+
+                    this.Log().Info("Writing files to app directory: {0}", target.FullName);
+                    await ReleasePackage.ExtractZipForInstall(
+                        Path.Combine(updateInfo.PackageDirectory, release.Filename),
+                        target.FullName,
+                        rootAppDirectory,
+                        progressCallback);
+
+                    // Delete the .not-finished file after extraction is completed
+                    if (File.Exists(notFinishedFilePath)) {
+                        File.Delete(notFinishedFilePath);
+                    }
 
                     return target.FullName;
                 });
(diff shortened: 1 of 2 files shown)


anaisbetts:
While this isn't the worst idea on its face, unfortunately solutions like this will cause 3rd party Antivirus to flag because you are moving executables out of a temp directory (generally the more you move an executable once it is written out, the more you are going to trigger them). 

You should not quit the app while updates are running - when the user selects Quit, don't kill the process, just close the window, await the update to finish, then quit

anaisbetts:
Instead of doing this, write a `.not-finished` file in the app folder, and delete it when the update successfully completes. Then, make the stubs ignore folders with `.not-finished` and delete any folders on update check that _do_ have `.not-finished`. This seems more complicated but it avoids moving executables around or renaming folders, which often triggers 3rd party antivirus

Also, don't quit your application while updates are running - instead of exiting, hide the window, await the update, then close once it finishes

DamonYu6:
> Instead of doing this, write a `.not-finished` file in the app folder, and delete it when the update successfully completes. Then, make the stubs ignore folders with `.not-finished` and delete any folders on update check that _do_ have `.not-finished`. This seems more complicated but it avoids moving executables around or renaming folders, which often triggers 3rd party antivirus

Agreed, I have submitted another commit, plz help review it, thanks. About `delete any folders on update check that do have .not-finished`, I don't need to add extra logic to handle this part since it already covers in https://github.com/Squirrel/Squirrel.Windows/blob/0ee9447d0f70744ad3624de953fc0eccf600990a/src/Squirrel/UpdateManager.ApplyReleases.cs#L306
and
https://github.com/Squirrel/Squirrel.Windows/blob/0ee9447d0f70744ad3624de953fc0eccf600990a/src/Squirrel/UpdateManager.ApplyReleases.cs#L96


> Also, don't quit your application while updates are running - instead of exiting, hide the window, await the update, then close once it finishes

Yes, during the upgrade process, we don't recommend users to exit the app. But we can't guarantee whether the app will be exited due to other reasons, such as crash or system shutdown. So we need this part logic to make sure users won't be able to launch the app due to the uncompleted upgrade.

src/Squirrel/UpdateManager.ApplyReleases.cs
```diff
@@ -308,13 +308,22 @@ Task<string> installPackageToAppDir(UpdateInfo updateInfo, ReleaseEntry release,
...
+
                     this.Log().Info("Writing files to app directory: {0}", target.FullName);
                     await ReleasePackage.ExtractZipForInstall(
                         Path.Combine(updateInfo.PackageDirectory, release.Filename),
                         target.FullName,
                         rootAppDirectory,
                         progressCallback);
 
+                    // Delete the .not-finished file after extraction is completed
+                    if (File.Exists(notFinishedFilePath)) {
```
anaisbetts: As a general programming rule, don't `stat` / `Exists` before doing a filesystem operation because it's a race condition, just Do It and deal with the failure

src/StubExecutable/StubExecutable.cpp
```diff
@@ -67,6 +72,13 @@ std::wstring FindLatestAppDir()
 
 		version::Semver200_version thisVer(s);
 
+		// Skip the directory which contains a .not-finished file
+		std::wstring appFolder = fileInfo.cFileName;
+        std::wstring dirPath = ourDir.substr(0, ourDir.size() - 5) + appFolder;
+        if (FileExists(dirPath + L"\\.not-finished")) {
```
anaisbetts: Indentation is weird?

anaisbetts (changes requested):
A few small things but this looks good

src/Squirrel/UpdateManager.ApplyReleases.cs
DamonYu6: fixed

src/StubExecutable/StubExecutable.cpp
DamonYu6: fixed

DamonYu6:
> A few small things but this looks good

Hi @anaisbetts , plz review it again, thx

anaisbetts (approved):

caesay:
Not sure if I missed something here, but I don't think this covers the `Update.exe --processStart` case, which will seemingly still launch the broken directory.

DamonYuRC:
@caesay In fact, when launching the app through "Update.exe --processStart," it initiates the latest version recorded in the local RELEASES file. However, if the last upgrade attempt failed, the RELEASES file remains unchanged, and as a result, it won't launch the broken directory.

PR #1864, opened 2023-09-21
Clarify --silent disabling first run

jcaron23:
Clarify that --silent also disables starting the app after install.


anaisbetts:
I think this is more appropriate to put in the documentation, I feel like this is the expected behavior for most people

PR #1868, opened 2023-11-23
Allow running as admin in Wine

### LINKED ISSUE #1720: feature request: Add flag to allow for bypassing UAC / admin check

foresto:
The admin privileges check introduced in #254 prevents working at all in Wine, which always reports admin privileges. Since there's no danger of causing permissions problems in that environment, allow it as a special case.

Hopefully closes #1720.

### DIFF
--- a/src/Setup/UpdateRunner.cpp
+++ b/src/Setup/UpdateRunner.cpp
@@ -34,6 +34,20 @@ void CUpdateRunner::DisplayErrorMessage(CString& errorMessage, wchar_t* logFile)
 	}
 }
 
+HRESULT CUpdateRunner::AreWeInWine()
+{
+	// NB: Behaving differently in Wine is *usually* discouraged
+	// https://wiki.winehq.org/Developer_FAQ#How_can_I_detect_Wine.3F
+	HMODULE hntdll = GetModuleHandle(L"ntdll.dll");
+	if (!hntdll) {
+		return S_FALSE;
+	}
+	if (!GetProcAddress(hntdll, L"wine_get_version")) {
+		return S_FALSE;
+	}
+	return S_OK;
+}
+
 HRESULT CUpdateRunner::AreWeUACElevated()
 {
 	HANDLE hProcess = GetCurrentProcess();
--- a/src/Setup/winmain.cpp
+++ b/src/Setup/winmain.cpp
@@ -110,7 +110,8 @@ int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
 
 	// If we're UAC-elevated, we shouldn't be because it will give us permissions
 	// problems later. Just silently rerun ourselves.
-	if (weAreUACElevated) {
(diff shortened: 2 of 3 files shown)


anaisbetts:
Flawless PR

src/Setup/UpdateRunner.cpp
```diff
@@ -34,6 +34,20 @@ void CUpdateRunner::DisplayErrorMessage(CString& errorMessage, wchar_t* logFile)
...
 
+HRESULT CUpdateRunner::AreWeInWine()
+{
+	// NB: Behaving differently in Wine is *usually* discouraged
+	// https://wiki.winehq.org/Developer_FAQ#How_can_I_detect_Wine.3F
+	HMODULE hntdll = GetModuleHandle("ntdll.dll");
+	if (!hntdll) {
+		return S_FALSE;
+	}
+	if (!GetProcAddress(hntdll, "wine_get_version")) {
```
anaisbetts: Needs to be L"wine_get_version"

src/Setup/winmain.cpp
```diff
@@ -110,7 +110,8 @@ int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
 
 	// If we're UAC-elevated, we shouldn't be because it will give us permissions
 	// problems later. Just silently rerun ourselves.
-	if (weAreUACElevated) {
+	// (Skip this check in Wine, which always reports admin privileges)
+	if (weAreUACElevated && CUpdateRunner::AreWeInWine() == S_FALSE) {
```
anaisbetts: Can we rewrite this condition to be `!= S_OK`?

anaisbetts (changes requested):

foresto:
Changes applied.

foresto:
Additional change applied: `L"ntdll.dll"`

anaisbetts:
Fixed this up manually and merged as https://github.com/Squirrel/Squirrel.Windows/commit/aaec357835514af347d6507d9127eda7eafbd740

PR #1873, opened 2024-01-26
fix multiple start errors

dansadley:
During setup, if a user performs mutliple extra mouse clicks while activating the installer, it will result in an error because the code is leaving the other instances of itself alive when performing the install. The result is it tries to burn itself to the ground and the file is locked due to being in use by the other process. It subsequently throws unhandled exception errors.

This change will check the processes if there is more than one of itself running and skip the entire install task after trying to log the event.


dansadley:
I am making further changes...

PR #1903, opened 2025-03-15
feat: skip generating _ExecutionStub.exe if it already exists

beyondkmp:
When _ExecutionStub.exe already exists in the files, avoid overwriting it. This is primarily because during the build process, _ExecutionStub.exe is automatically generated with my custom complex signatures process(Information is written to _ExecutionStub.exe through WriteZipToSetup, which only adds custom complex signatures), but Squirrel Windows overwrites it, causing the loss of signatures.


src/build_official.cmd
t3chguy, line 1: This seems to be entirely whitespace changes, is that intentional?

beyondkmp: Yes, I changed \r to \r\n because it was causing errors on my machine.

```
The system cannot find the batch label specified - StartDeveloperCommandPrompt.
```

However, since it works fine on GitHub Actions, I reverted the change.

PR #1916, opened 2026-04-09
[FE-7847] Change OpenSetupLog Target

rickstergg:
We write to this file, so we should open it when users have a failed install


rickstergg:
Apologies, meant to open this up against our fork, don't mind me! :)

